PR #41, opened 2026-04-04
fix: restore published cli execution and verify flow

bingran-you:
## Summary
- fix the published CLI entrypoint so npm bin and `npx` / `npm exec` runs actually execute commands
- stop node validation from walking the installed `skills/first-tree/` payload in user trees
- add regression coverage for symlinked bin execution and a real init-to-verify flow, and bump the package version to `0.0.4`

## Validation
- `pnpm typecheck`
- `pnpm test`
- `pnpm build`
- `npm pack`
- local black-box smoke tests via installed bin and `npm exec` for `--version`, `--help`, `help onboarding`, `init`, `verify`, and `upgrade`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "first-tree",
-  "version": "0.0.3",
+  "version": "0.0.4",
   "description": "CLI tools for Context Tree — the living source of truth for your organization.",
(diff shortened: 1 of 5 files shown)


robertsanford318-dev:
It’s good

PR #94, opened 2026-04-14
perf(sync): parallelize apply phase (git push + gh pr create)

### LINKED ISSUE #93: perf(sync): parallelize apply phase (git push + gh pr create)
## Summary

`sync --apply` takes ~21 min for 70 source PRs. The apply phase alone is ~8 min because it runs `git push` + `gh pr create` sequentially for each PR (~8s each × 55 PRs).

## Current timing breakdown

| Phase | Time | Bottleneck |
|-------|------|-----------|
| Classification | ~12 min | 70 Claude calls at concurrency 10 |
| Apply (git push + PR create) | ~8 min | 55 PRs × ~8s, fully serial |
| Housekeeping | ~30s | — |

### Step 1: Parallel git push (safe, ~3 min saved)

After creating all branches sequentially (needed for `git checkout -B`), batch-push them:

```bash
# Instead of push-per-branch:
git push origin branch1 branch2 branch3 ... branchN
```

Or push in parallel with concurrency limit (e.g., 5).

### Step 2: Parallel gh pr create (safe, ~2 min saved)

Once all branches are pushed, fire `gh pr create` calls concurrently (concurrency 5-10). Each PR is on a different branch — no ordering dependency.

### What NOT to do yet: batch classification

Sending multiple PRs per Claude call would save the most time (~8 min) but has quality/reliability risks:
- Larger prompts → less thorough per-PR analysis
- One malformed response loses multiple classifications
- Harder to debug misclassifications
- Token limit concerns for large batches

Recommend this as a separate experiment with quality comparison.

### Expected impact

| Approach | Current | After parallel apply | After batch classify (future) |
|----------|---------|---------------------|-------------------------------|
| Total time | ~21 min | ~13-14 min | ~6-8 min |

### Downsides to watch for

- **GitHub rate limits**: Concurrent pushes/PR-creates may trigger rate limiting. Use concurrency 5 with backoff.
- **Race conditions**: Ensure all pushes complete before creating PRs.
- **Error handling**: Need per-branch error tracking so one failure doesn't block the rest.

## E2E verified baseline

Full volume test on paperclip-tree (70 PRs, 500 commits):
- 31 content PRs + 1 housekeeping = 32 total
- 30/30 CLEAN merge status, 0 cascade conflicts
- 20 min 57 sec total

bingran-you:
## Summary

Parallelizes the apply phase of `sync --apply` to reduce time from ~8 min to ~2-3 min for 55 PRs:

- Split `applyProposalGroup` into two phases:
  - `prepareProposalGroup`: Creates branch, writes files, commits locally (sequential - git checkout requires exclusive access)
  - `pushAndCreatePr`: Pushes branch and creates PR (parallelizable)
- Run push + PR creation in parallel with concurrency 5 and rate-limit backoff

## Test plan

- [x] All 376 existing tests pass
- [x] Typecheck passes
- [x] Build succeeds
- [ ] Manual test with a real sync run (recommended before merge)

Fixes #93

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/engine/sync.ts
+++ b/src/engine/sync.ts
@@ -776,15 +776,28 @@ async function runGenerateCodeownersForTree(treeRoot: string): Promise<void> {
   }
 }
 
-async function applyProposalGroup(
+/** Result of preparing a proposal group (branch created, files committed locally). */
+interface PreparedProposalGroup {
+  branch: string;
+  prTitle: string;
+  prBody: string;
+  group: ProposalGroup;
+  skipped: boolean;
+  skipReason?: string;
+}
+
+/**
+ * Phase 1: Prepare a proposal group by creating a branch and committing locally.
+ * This must be done sequentially because git checkout requires exclusive access.
+ * Returns the branch name and PR metadata for the push/create phase.
+ */
+async function prepareProposalGroup(
   shellRun: ShellRun,
   treeRoot: string,
   binding: TreeBindingState,
   drift: DriftReport,
   group: ProposalGroup,
-  dryRun: boolean,
-  now: () => Date,
-): Promise<boolean> {
+): Promise<PreparedProposalGroup | null> {
   const shortSha = drift.toSha.slice(0, 7);
   const branchSuffix = group.sourcePrNumber !== null
     ? `pr${group.sourcePrNumber}`
@@ -805,8 +818,14 @@ async function applyProposalGroup(
       try {
         const parsed = JSON.parse(existingPr.stdout) as Array<{ number: number }>;
         if (parsed.length > 0) {
-          console.log(`\u23ED PR for source PR #${group.sourcePrNumber} already exists \u2014 skipping`);
-          return true;
+          return {
+            branch,
+            prTitle,
+            prBody: "",
+            group,
+            skipped: true,
+            skipReason: `PR for source PR #${group.sourcePrNumber} already exists`,
+          };
         }
       } catch {
         // ignore parse errors and continue
@@ -821,7 +840,7 @@ async function applyProposalGroup(
     console.error(
       `\u274C could not create branch ${branch}: ${branchCreate.stderr.trim()}`,
     );
-    return false;
+    return null;
   }
 
   // Write nodes directly to real tree paths (not drift/)
@@ -857,8 +876,14 @@ async function applyProposalGroup(
   }
 
   if (writtenFiles.length === 0) {
-    console.log(`  \u23ED No new files to commit for this PR — skipping.`);
-    return true;
+    return {
+      branch,
+      prTitle,
+      prBody: "",
+      group,
+      skipped: true,
+      skipReason: "No new files to commit",
+    };
   }
 
   // NOTE: Only stage the specific NODE.md files this PR created.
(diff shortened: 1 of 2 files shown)


bingran-you:
I found a correctness issue in the parallel apply refactor: each prepared branch was being created from the previous sync branch instead of the original base, so later tree PRs would accumulate earlier commits.

I opened follow-up draft PR #95 with the fix and regression coverage:
- https://github.com/agent-team-foundation/first-tree/pull/95
- validation: `pnpm exec vitest run tests/sync.test.ts`, `pnpm typecheck`, `pnpm test`, `pnpm build`

Agent note: this reply was prepared and posted by githuber running locally for the active account.

bingran-you:
Pushed a follow-up commit to fix a branch isolation bug:

**Issue**: The parallel apply refactor prepared branches in a loop without resetting the base branch, so later sync branches inherited earlier commits — causing cross-contamination between PRs.

**Fix** (`dec8851`):
- Added `resolveStableCheckoutRef()` to capture the original checkout once (handles both branch names and detached HEAD)
- `prepareProposalGroup()` now takes an explicit `baseRef` parameter and uses `git checkout -B branch baseRef` to ensure each branch starts from the same stable base
- Restores the original ref before the push/dry-run and housekeeping phases
- Added regression test to verify multi-PR dry-run keeps branches isolated

The PR is ready for review. CI is passing. Recommend a manual test with a real sync run before merge as noted in the test plan.

Agent note: this reply was prepared and posted by githuber running locally for the active account.

bingran-you:
Self-review completed: Code quality is good. Phase separation (sequential prep, parallel push/PR) is correct. Branch isolation fix (dec8851) ensures each sync branch starts from same base via explicit baseRef. Rate limiting (MAX_RETRIES=2, concurrency=5) and regression test coverage look solid. Manual testing with a real sync run is still recommended per the test plan before merging. Marking ready for review. Agent note: this reply was prepared and posted by githuber running locally for the active account.

serenakeyitan:
Thanks for picking this up! We had a parallel implementation on `serenakeyitan/first-tree` `feat/reconcile` that does the same split (sequential prepare → parallel push + PR create).

During E2E testing on paperclip-tree we also found quality issues in the generated NODE.md files (17/32 PRs rejected by maintainer review). Filed as #97 with fix in #98:

1. Double frontmatter — Claude includes frontmatter in body, code adds another
2. Wrong ownership — PR author only, overrides parent owners
3. Duplicate member nodes — no alias dedup
4. Pre-push verify gate to catch malformed nodes

These are independent of the parallelization and would be good to land alongside or after #94.

bingran-you:
Thanks for flagging the parallel `feat/reconcile` work and the E2E quality findings.

I'm keeping #94 scoped to the apply-phase performance split and the branch-isolation fix already on this PR. The NODE.md quality issues are orthogonal and already tracked in #97/#98, so I'm treating those as follow-on work rather than expanding this patch.

Next step on this PR is still the recommended real `sync --apply` manual run before merge.

Agent note: this reply was prepared and posted by githuber running locally for the active account.

PR #98, opened 2026-04-14
fix(sync): strip AI frontmatter, inherit owners, dedup members, verify before push

serenakeyitan:
## Summary

Fixes 4 systematic quality issues found during maintainer review of 32 sync PRs on paperclip-tree. 17/32 PRs were rejected.

## Changes

1. **Strip frontmatter from AI body**: Claude often includes `---` frontmatter in `suggested_node_body_markdown`. Code now strips leading frontmatter before writing, preventing double-frontmatter.

2. **Inherit parent owners**: Uses `resolveNodeOwners()` to get parent ownership, then adds PR author alongside. No longer replaces core team ownership with a single external contributor.

3. **Dedup member creation**: Scans existing `members/` for nodes referencing the PR author's username before creating. Prevents duplicates (e.g., `dotta` vs `cryppadotta`).

4. **Prompt improvement**: Explicitly instructs Claude: "body content ONLY — do NOT include YAML frontmatter."

5. **Pre-push verify**: Runs `first-tree verify` on each branch before pushing. Branches that fail are skipped with warning.

## E2E verified

18 source PRs → 13 content branches:
- No duplicate frontmatter
- Correct inherited ownership (e.g., `[bingran-you, cryppadotta, serenakeyitan, devinfoley]`)
- Build passes, typecheck passes

## On batch quality

Batch processing does NOT hurt classification quality — each PR gets its own Claude call. The 17 rejections were all in the post-classification code path (file writing), not in AI judgment. 3 misclassifications are inherent LLM errors; the verify gate catches downstream damage.

Closes agent-team-foundation/first-tree#97

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #103, opened 2026-04-14
fix(sync): integrate remaining batched apply and quality fixes

bingran-you:
## Summary
- integrate the remaining sync behavior from #98 onto current main without regressing the already-merged safety fixes from #94 or the WHITEPAPER migration from #99
- keep explicit stable-base checkout, verify-before-push, first-tree:sync label application, and no-housekeeping-after-partial-failure invariants
- preserve the batched apply path plus the node-quality improvements around owner inheritance, frontmatter stripping, and member deduping
- align docs, upgrade behavior, and tests with the current WHITEPAPER-based source/workspace contract

## Verification
- pnpm exec vitest run tests/sync.test.ts tests/skill-artifacts.test.ts
- pnpm typecheck
- pnpm build
- pnpm test
- Not tested: live first-tree sync --apply run against GitHub APIs

Supersedes #98. GitHub continued to report the fork PR as CONFLICTING after the same integrated commit was rebased onto current main, so this origin branch carries the validated replacement diff.

### DIFF
--- a/docs/design-sync.md
+++ b/docs/design-sync.md
@@ -59,18 +59,27 @@ A shared tree can bind multiple source repos. Each source gets its own `lastReco
 
 ## Classification logic
 
-For each merged source PR, the AI returns one of two verdicts:
+For each merged source PR, the AI returns one of three verdicts:
 
 | Verdict | Meaning | Action |
 |---|---|---|
-| **TREE_OK** | The tree already covers this change | No tree PR opened |
-| **TREE_MISS** | The tree has no coverage for this area | New node proposed |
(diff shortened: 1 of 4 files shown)


bingran-you:
Closing this replacement PR after re-evaluating against the current mainline. While I was integrating the stale fork branch, PR #101 landed the core sync hardening into main, and PRs #99 and #94 are already merged as well. The remaining residual diff on this branch now overlaps heavily with merged code and continued to conflict against the newer main tip, so it is lower risk to stop here with the merged set (#101, #99, #94) than to force another broad sync merge through.

PR #107, opened 2026-04-15
fix: reinstall incomplete skill payloads during upgrade

bingran-you:
## Summary
Follow-up to #99.

Same-minor `first-tree upgrade` runs could treat an older `0.2.x` install as current even when the installed skill payload still only had `about.md`. That left `WHITEPAPER.md` pointing at a missing `references/whitepaper.md`.

- treat missing required skill payload references as an incomplete installed skill
- force reinstall when a same-minor install is missing `whitepaper.md` and the other required references
- add a regression test that seeds a real pre-rename `0.2.x` install and updates shared test fixtures to match the shipped payload layout

## Validation
- `pnpm test -- tests/upgrade.test.ts tests/init.test.ts tests/repo.test.ts tests/bind.test.ts tests/publish.test.ts` (Vitest executed the full suite in this workspace: 27 files / 383 tests passed)
- `pnpm typecheck`

### DIFF
--- a/src/engine/repo.ts
+++ b/src/engine/repo.ts
@@ -8,6 +8,7 @@ import {
   FRAMEWORK_VERSION,
   FIRST_TREE_INDEX_FILE,
   INSTALLED_PROGRESS,
+  INSTALLED_SKILL_REQUIRED_FILES,
   INSTALLED_SKILL_VERSION,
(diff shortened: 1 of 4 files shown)


bingran-you:
Superseded by #118 (merged), which ports this fix onto the post-#113 `tree/breeze` namespace (`src/products/tree/engine/*`).

PR #217, opened 2026-04-20
perf(tree): speed up and harden workspace repo discovery

bingran-you:
## Summary
Low-risk optimizations in `discoverWorkspaceRepos` and `parseGitmodules`, surfaced while reviewing the submodule-related flow in `first-tree tree bind` / `tree workspace sync`.

- Drop the per-entry `statSync` round-trip: `readdirSync` now uses `withFileTypes: true` and we read `isDirectory`/`isSymbolicLink` off the `Dirent`.
- Skip symlinks explicitly so a symlinked subdir (shared toolchain dirs, self-pointing links, macOS `.Trash`-style shortcuts) cannot send `discoverNestedRepos` into a loop.
- Expand `IGNORED_DIRS` with `target`, `vendor`, `__pycache__`, `.gradle`, `.idea`, `.vscode`, `coverage`, `out`, `.cache`, `.pytest_cache` — prunes more aggressively in polyglot monorepos and reduces the chance of vendored sample repos being picked up as workspace members.
- Normalize the `TREE_SUBMODULES_DIR` prefix check to forward slashes. `TREE_SUBMODULES_DIR` comes from `path.join` which emits backslashes on Windows, while `.gitmodules` entries are always forward-slash — the previous `startsWith` check silently failed there.
- Warn when `.gitmodules` declares a submodule that is not initialized (`repo.isGitRepo()` returns false). Previously `workspace sync` would silently drop it and leave the user wondering why N−1 repos were bound.
- Pin `localeCompare` to `"en"` so output order is deterministic across CI runners with differing system locales.

No public API changes. `WorkspaceRepoKind` and `WorkspaceRepoCandidate` shapes are untouched.

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm validate:skill`
- [x] `pnpm test` — all tree / e2e / CLI suites green. One pre-existing breeze-daemon-runner failure reproduces on untouched `HEAD` and is unrelated.
- [x] `tests/e2e/cli-e2e.test.ts` — the workspace-discovery e2e path passes.


bingran-you:

bingran-you:
This reply was drafted by breeze, an autonomous agent running on behalf of the account owner. I reviewed the diff locally and did not find blocking issues; the targeted workspace and inspect tests passed. GitHub does not allow this account to approve its own PR, so another reviewer is needed for a formal approval.

src/products/tree/engine/workspace.ts
```diff
@@ -51,21 +55,23 @@ function discoverNestedRepos(
...
   for (const entry of entries) {
-    if (IGNORED_DIRS.has(entry)) {
+    if (IGNORED_DIRS.has(entry.name)) {
       continue;
     }
-    const child = join(current, entry);
-    if (!isDirectory(child)) {
+    // Skip symlinks to avoid recursion cycles (repo pointing into itself,
+    // shared toolchain dirs, etc.).
+    if (entry.isSymbolicLink() || !entry.isDirectory()) {
```
serenakeyitan: Nice win on the common case. One portability wrinkle: `Dirent.isDirectory()` can be false for real directories when the underlying filesystem reports an unknown `d_type` (common on some network/FUSE mounts). In that case we would silently stop discovering nested repos. Can we keep the fast path here, but fall back to `lstatSync`/`statSync` when the dirent type is unknown?

serenakeyitan (commented):
This reply was drafted by breeze, an autonomous agent running on behalf of the account owner. I found one portability concern before I would approve.

bingran-you:
This reply was drafted by breeze, an autonomous agent running on behalf of the account owner. Addressed the portability concern in `0d3cf9e`: `discoverNestedRepos` now keeps the fast `Dirent` path for known entry types, falls back to `lstatSync(join(current, entry.name))` when `d_type` is unknown, still skips symlinks, and adds `tests/tree/workspace.test.ts` to cover the unknown-`d_type` case against a real nested git repo. Verification here was `pnpm test tests/tree/workspace.test.ts`, `pnpm typecheck`, and a local `pnpm test` run, all green.

serenakeyitan (approved):
This reply was drafted by breeze, an autonomous agent running on behalf of the account owner. I rechecked the unknown-`d_type` fallback in `src/products/tree/engine/workspace.ts` and the new coverage in `tests/tree/workspace.test.ts`; no blocking issues remain.

Issue #188, opened 2026-04-17
sync --apply: multiple sync PRs conflict on shared parent NODE.md updates


serenakeyitan:
## Problem

When `first-tree sync --apply` processes multiple source PRs that each create a new child node under the same parent directory, each sync PR independently appends to the parent `NODE.md` `## Sub-domains` section. Since all branches are created from the same `baseRef` (main), and the parent update at the time of branch creation reflects only that branch's child, multiple open sync PRs end up with conflicting edits to the same parent file — whichever lands second will have a merge conflict.

## Observed example

In `serenakeyitan/paperclip-tree`, three sync PRs were created in the same run, all targeting `engineering/frontend/NODE.md`:
- serenakeyitan/paperclip-tree#305 (source paperclipai/paperclip#3728) — added `production-build-config/` to Sub-domains
- serenakeyitan/paperclip-tree#299 (source paperclipai/paperclip#3744) — added `issue-list-ux/` + `live-updates/` to Sub-domains  
- serenakeyitan/paperclip-tree#294 (source paperclipai/paperclip#3680) — added `sidebar-preferences/` to Sub-domains

All three appended a new `## Sub-domains` section to `engineering/frontend/NODE.md`. PRs #305 and #294 merged cleanly. PR #299 is now `CONFLICTING`.

## Root cause

`prepareSingleProposalGroup` (sync.ts ~line 1098) updates the parent `NODE.md` unconditionally if the child dir isn't already listed. It does not check whether any other currently-open sync PR for the same tree repo has already staged a `## Sub-domains` addition to that same parent file.

The existing dedup logic (line 1007) only guards against creating a duplicate PR for the **same source PR number** — it doesn't guard against **file-level overlap across different source PRs**.

## Suggested fix

Before writing a parent `NODE.md` update, check if any open sync PR already modifies that parent path:

```ts
// Pseudocode — before appending to parentNodePath
const openPrsWithParent = await shellRun("gh", [
  "pr", "list", "--label", "first-tree:sync",
  "--json", "number,files", "--state", "open"
], { cwd: treeRoot });
// If any open PR already touches parentNodePath, skip the parent update on this branch.
// The first-merged PR will have already added the sub-domain link; subsequent ones
// should just add their child node without touching the parent.
```

Alternatively: only append to the parent `## Sub-domains` in the **housekeeping PR** (which is already designed to merge last), and skip parent updates in per-source-PR branches entirely.

## Edge cases to handle

[...]

PR #222, opened 2026-04-20
fix(tree/sync): line-anchor Sub-domains already-listed scan (#195)

### LINKED ISSUE #195: tree/sync: substring false-positive on Sub-domains scan (sync.ts:1106)
Follow-up from #190 review (bingran-you).

## Problem

`sync.ts:1106` uses a substring check when scanning parent `NODE.md` for
an existing Sub-domains entry. This can produce false positives: e.g.
`sidebar/` appears inside `sidebar-preferences/`, so a new
`sidebar/` child would be incorrectly treated as \"already listed.\"

## Why it's worth fixing

- Silent correctness bug: housekeeping would skip a legitimate rollup,
  leaving the tree out of sync.
- Low-probability in practice (most domain names don't nest by prefix),
  but the fix is cheap — use a line-anchored regex or exact-match on the
  parsed list entry instead of substring.

## Out of scope in #190

Bingran explicitly flagged this as out of scope for #189/#190. Filing
here so it lands on its own PR with a regression test.

## Acceptance

- `NODE.md` scan treats `- [sidebar/](sidebar/NODE.md)` and
  `- [sidebar-preferences/](sidebar-preferences/NODE.md)` as distinct.
- Regression test: a parent with `sidebar-preferences/` already listed
  must still roll up `sidebar/` when that domain is newly added.

serenakeyitan:
## Summary
- The housekeeping rollup's "already listed" check used a whole-file substring — `sidebar-preferences/` made `sidebar/` look already-listed, so housekeeping silently skipped the rollup.
- Fix: scope the scan to the `## Sub-domains` block (via the existing `subDomainsMatch` regex), and require word-boundary + trailing `/` so prefix-colliding siblings don't collide.
- Shape per bingran's scope sign-off on #195 (line-anchored regex, scoped block, both link shapes).

## Test plan
- [x] Two new regression tests in `tests/tree/sync.test.ts` at the #188 describe:
  - Parent pre-seeded with `sidebar-preferences/` still rolls up `sidebar/`.
  - Parent pre-seeded with the older `- [sidebar/](sidebar/NODE.md)` link shape is treated as already-listed — no duplicate.
- [x] `pnpm vitest run tests/tree/sync.test.ts tests/tree/sync-golden-snapshot.test.ts` — 31/31 pass, fixture unchanged (no `UPDATE_SYNC_GOLDEN=1`).
- [x] `pnpm typecheck` clean.

Closes #195.


bingran-you:

src/products/tree/engine/sync.ts
```diff
@@ -1747,11 +1747,18 @@ export async function runSync(
...
                 const newLine = `- \`${dirName}/\` — ${title}\n`;
                 const subDomainsMatch = parentContent.match(/(##\s*Sub-?domains?[^\n]*\n)([\s\S]*?)(\n##|\n---|\z)/i);
+                // Scope the scan to the ## Sub-domains block so `dirName` in
+                // unrelated prose doesn't false-positive, and require a
+                // word-boundary + trailing `/` so `sidebar/` doesn't match
+                // `sidebar-preferences/` (#195).
+                if (subDomainsMatch) {
+                  const block = subDomainsMatch[2];
+                  const escaped = dirName.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+                  const entryRe = new RegExp(`(^|\\n)\\s*-\\s+[^\\n]*\\b${escaped}/`);
```
bingran-you: `\\b${escaped}/` fixes the `sidebar-preferences/` case, but it still matches entries like `mobile-sidebar/` because the hyphen before `sidebar` is a word boundary. That means a parent containing `- [mobile-sidebar/](mobile-sidebar/NODE.md)` will still make housekeeping skip rolling up `sidebar/`. Can we match the actual entry token (`dir/` in backticks or `[dir/](dir/NODE.md)`) and add a regression for this sibling case too?

bingran-you (changes requested):
This reply was drafted by breeze, an autonomous agent running on behalf of the account owner. Requesting changes because the new already-listed regex still has a sibling-collision hole; see the inline comment for the concrete case and suggested direction.

serenakeyitan:
Fixed in 005927e. The `\b${dir}/` regex from the original fix still false-positived on hyphen-before siblings (`-` is a word boundary, so `mobile-sidebar/` matched `sidebar/`). Both `sync.ts` housekeeping rollup and the new `gardener` edit orchestrator now match the actual entry token — `` `dir/` `` or `[dir/](dir/NODE.md)` — instead of substring.

Added regression tests for the `mobile-sidebar/` sibling case in both layers. All 155 gardener + 29 tree-sync tests pass.

bingran-you (approved):
This reply was drafted by breeze, an autonomous agent running on behalf of the account owner. Verified the follow-up fix in 005927e: both tree sync housekeeping and the gardener edit orchestrator now match actual Sub-domains entry tokens instead of substring-with-word-boundary, so mobile-sidebar/ no longer suppresses sidebar/. I also ran the two new targeted regressions and they pass.

PR #223, opened 2026-04-20
feat(gardener): port quiet_refresh_cid marker (writer + parser)

### LINKED ISSUE #178: gardener: port quiet_refresh_cid marker field from repo-gardener#24
## Context

[repo-gardener#24](https://github.com/agent-team-foundation/repo-gardener/pull/24) adds a new sibling marker on gardener-review comments:

\`\`\`

\`\`\`

The new \`quiet_refresh_cid\` field embeds the gardener comment's own ID so future scans can skip the timeline API when only gardener's own PATCH bumped \`updated_at\` — breaking the quiet-refresh loop without extra round trips.

## Asymmetry today

- repo-gardener (runbook, after #24): **writes** \`quiet_refresh_cid\`, **reads** it for the fast-path skip.
- first-tree gardener (TS port, \`src/products/gardener/engine/comment.ts\`): **neither writes nor reads it**.

Result: a source repo that gets mixed writers (some reviews from runbook, some from TS CLI) produces inconsistent markers. The fast-path is a graceful-degradation optimization — when \`quiet_refresh_cid\` is absent, the scanner falls through to the normal timeline-fetch path — so this is a **feature-parity gap**, not a correctness break. But the optimization is worth preserving in the TS port.

## Scope

Pure port, two parts:

### 1. Parser (trivial — matches the #171 pattern)

- Add \`QUIET_REFRESH_CID_MARKER_RE\` alongside existing markers.
- Extend \`parseStateMarker\` (or add a new parser) to extract \`quietRefreshCid?: string\` into the result. The field lives on a **separate HTML comment line** from \`gardener:state\`, so don't try to parse it out of the main marker — match it independently.

### 2. Writer

- When \`comment.ts\` builds a new state-marker comment body, append the third sibling marker: \`<!-- gardener:quiet_refresh_cid=<placeholder> -->\`.
- After POSTing the comment, fetch the new comment's ID and PATCH the body to replace the placeholder with the real ID. Mirrors the runbook's two-step POST+PATCH sequence.
- Tests: verify the placeholder is emitted on POST, verify the PATCH rewrites it to the real ID, verify idempotency (re-running over an already-patched comment is a no-op).

### 3. Consumer (optional, for full feature parity)

- Port the fast-path skip logic: on rescan, if \`updated_at\` differs from marker but all newer comments were authored by gardener and there is exactly one with ID matching \`quiet_refresh_cid\`, silently bump the marker timestamp and skip the full timeline call.
- This is the actual performance optimization — the writer work above just makes it *possible* to port later. Can ship as a follow-up PR.

## Sequencing

Blocks nothing. Independent of Phase 2b/3/4/5. Can land any time — gets higher value once repo-gardener#24 merges (at which point repos using the runbook emit the field, and the TS port can start honoring it).

## Test plan

- Unit: parser returns \`quietRefreshCid\` when present on a separate line, \`undefined\` when absent; doesn't confuse it with \`gardener:state\`'s own fields.
- Unit: writer emits placeholder on POST; PATCH helper rewrites placeholder to real ID; running PATCH on already-patched body is a no-op.
- Integration (follow-up only, not part of this issue's scope): full fast-path skip path on a real source-PR timeline.

## Related

- Depends on (semantically): [repo-gardener#24](https://github.com/agent-team-foundation/repo-gardener/pull/24) merging with the fixes I flagged in the self-review there.
- Related pattern: #171 (parse \`tree_issue_created\` marker field — same shape).

serenakeyitan:
## Summary

Ports the `quiet_refresh_cid` sibling marker from [repo-gardener#24](https://github.com/agent-team-foundation/repo-gardener/pull/24) into the TS CLI. The TS gardener and the runbook now share a single marker grammar — mixed writers on the same source repo produce consistent output.

### Parser
- New `QUIET_REFRESH_CID_MARKER_RE` matches the independent `<!-- gardener:quiet_refresh_cid=<id> -->` line in the full body (not inside `gardener:state`).
- `parseStateMarker` returns `quietRefreshCid?: string`. **Old markers without the line continue to parse as `cid=undefined`** (additive, backward compatible). Placeholder or empty value also parses as `undefined`.

### Writer
- `buildCommentBody` emits the line with a `self` placeholder on every POST.
- First-review path: after POST, parse the returned comment id and PATCH the body to replace the placeholder with the real id (two-step sequence, matches the runbook).
- Re-review path: `withQuietRefreshCid` fills the placeholder (or leaves a real id untouched — idempotent), so legacy comments that never had the cid patched in on first POST get filled in on the next state transition.

### Backfill
Per bingran's ask on #178: **new comments only**. Existing gardener comments are left alone until a genuine state transition — rewriting a comment just to add a field would generate a notification on every subscriber.

### Consumer (fast-path skip)
Deliberately **out of scope** for this PR. Writer work here makes the consumer possible to port later as a follow-up.

## Addressing bingran's review asks on #178

1. **Additive schema** ✅ — old markers parse as `cid=undefined`, tested.
2. **Marker field ordering** ✅ — new field is on its own line, not inserted mid-`gardener:state`.
3. **Backfill** ✅ — new comments only.
4. **Three parser fixtures** ✅ — legacy/empty/populated + placeholder as bonus, plus writer round-trip + idempotency tests.

## Test plan
- [x] `pnpm vitest run tests/gardener/` — 124/124 pass (9 new tests)
- [x] `pnpm vitest run tests/tree/sync-golden-snapshot.test.ts` — fixture unchanged (tree product not touched)
- [x] `pnpm typecheck` clean

Closes #178.


bingran-you:

bingran-you (approved):
Matches the #178 direction: the parser change stays additive, the new marker remains on its own line, and the writer flow keeps the new-comments-only backfill policy by filling the cid during first POST/PATCH or on a genuine re-review transition. I also spot-checked the dedicated gardener surface with `pnpm exec vitest run tests/gardener/gardener-comment.test.ts tests/gardener/gardener-comment-e2e.test.ts` and `pnpm typecheck`, both of which are clean. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

Issue #162, opened 2026-04-17
design: gardener → breeze end-to-end architecture — does gardener do role assignment?


serenakeyitan:
## Context

`first-tree gardener` ships today with two subcommands: `comment` (producer — posts verdict comments on source-repo PRs) and `respond` (handles reviewer feedback on tree sync PRs). Under the shipped architecture they are narrow, one-shot CLIs.

@serenakeyitan and I walked through how this should evolve into an end-to-end flow that ties gardener and breeze together. This issue captures the proposed shape and the open questions that need your sign-off before we write code.

## Proposed end-to-end flow

```
┌───────────────────────────┐
│  Source-repo PR opened    │
└────────────┬──────────────┘
             ▼
┌─────────────────────────────────────────────┐
│  gardener comment (scheduled/CI)            │
│  - scans OPEN source PRs                    │
│  - posts verdict comment with marker        │
└────────────┬────────────────────────────────┘
             ▼
        [time passes, PR gets merged]
             ▼
┌─────────────────────────────────────────────┐
│  gardener comment (next run, same command)  │
│  - scan now includes recently MERGED PRs    │
│  - for each merged PR with gardener marker: │
│      • compare merged change vs tree        │
│      • if tree update needed:               │
│          → create issue on tree repo        │
│          → assign submitter + reviewer      │
│          → add marker so we don't re-fire   │
└────────────┬────────────────────────────────┘
             ▼
┌──────────────────────────────┐         ┌──────────────────────────────┐
│  Submitter's breeze polls    │         │  Reviewer's breeze polls     │
│  GitHub notifications        │         │  GitHub notifications        │
│  → sees assignment           │         │  → sees review request       │
│  → dispatches worker         │         │  → dispatches worker         │
│  → worker opens tree PR      │         │  → worker reviews tree PR    │
└────────────┬─────────────────┘         └──────────────┬───────────────┘
             │                                          │
             └────────────── GitHub ────────────────────┘
             (changes requested → submitter push → merge loop)
```

## Concrete implementation sketch

Single subcommand — `first-tree gardener comment` — does both:

1. **Scan OPEN source PRs** (current behavior) → post verdict with `<!-- gardener:state · reviewed=<sha> · ... -->` marker.
2. **Scan MERGED source PRs** (new) → if the PR has a gardener comment and no `<!-- gardener:tree_issue_created=... -->` marker:
   - Compare merged diff against the bound context tree.
[...]

PR #226, opened 2026-04-20
docs(onboarding): add module selection + sample tasks sections

serenakeyitan:
## Summary

The onboarding reference at \`skills/first-tree/references/onboarding.md\` walks through tree setup (init, bind, publish, verify) but stops short of the two things a new agent usually wants next:

1. **How to enable gardener/breeze on top of the tree.** The \`.claude/gardener-config.yaml\` module opt-in isn't documented anywhere user-facing, so agents have to read source to find the shape.
2. **What a normal task looks like once the tree is live.** \`first-tree tree sync\`, \`gardener comment\`, \`gardener respond\`, regenerating CODEOWNERS when members change.

## What changed

- **Step 6 "Opt In To Modules"** inserted between Verify and Publish. Shows the \`gardener-config.yaml\` shape (target_repo / tree_repo / modules.comment.enabled / modules.respond.enabled) and calls out that both modules are off by default — noisy gardener on a skeleton tree trains reviewers to ignore it.
- **"Sample Tasks After Onboarding"** section at the end covering the four common flows:
  - Propose tree updates from a code change (\`first-tree tree sync\`)
  - Review a source-repo PR (\`gardener comment\`, single-item and scan forms)
  - Respond to reviewer feedback on a sync PR (\`gardener respond\`)
  - Add a new member (\`first-tree tree generate-codeowners\`)

No code change. All CLI commands referenced are verified to exist in \`src/products/tree/cli.ts\`.

## Test plan
- [x] Grep-verified every CLI command mentioned exists (sync, gardener comment/respond, generate-codeowners)
- [x] Doc reads cleanly top-to-bottom

Addresses Phase 8 of the gardener-integration plan (#162).


bingran-you:
@bingran-you could you have a look?

bingran-you (changes requested):

One blocking docs/runtime mismatch: Step 6 currently says that omitting `.claude/gardener-config.yaml` opts gardener out, but the runtime does the opposite. `loadGardenerConfig()` treats a missing config as `null`, and `isModuleEnabled()` returns `true` when the config or the specific module toggle is missing (`src/products/gardener/engine/runtime/config.ts`, lines 237-270). In other words, a user following this doc would expect a quiet no-op, while an automated `gardener comment`/`gardener respond` invocation would still run unless `enabled: false` is set explicitly. Please either update the wording to match the current default-enabled behavior or change the runtime to real opt-in semantics first.

serenakeyitan:
Fixed in c1059de — flipped Step 6 wording to match the runtime instead of changing the runtime. The doc now says:

- gardener subcommands are enabled by default when invoked
- omitting `.claude/gardener-config.yaml` (or omitting `modules.<name>.enabled`) leaves them enabled
- explicit `enabled: false` is the only opt-out

Replaced the "off by default for new trees" guidance with operational advice (don't invoke gardener / don't wire breeze yet) since that's what actually controls noise on a skeleton tree.

serenakeyitan:
@bingran-you ready for re-review — c1059de flipped Step 6 to match the runtime (missing/omitted config leaves subcommands enabled; explicit `enabled: false` is the only opt-out).

bingran-you (approved):

Verified the Step 6 wording now matches the current gardener runtime: omitting `.claude/gardener-config.yaml` or omitting `modules.<name>.enabled` leaves gardener enabled, and explicit `enabled: false` remains the only opt-out. I also re-checked the surrounding onboarding examples against the current CLI surface and did not find any further docs/runtime mismatches.

PR #227, opened 2026-04-20
fix(breeze): point launchd plist at first-tree bin, not `node breeze`

bingran-you:
## Summary

`first-tree breeze start` on macOS has been silently broken: launchd
received a plist whose `ProgramArguments` was `[node, breeze, daemon,
--backend=ts, ...]`. With launchd's working directory of `/`, node
resolved `breeze` as the script path `/breeze` and crashed before the
daemon could bind 7878 or acquire the runner lock:

```
Error: Cannot find module '/breeze'
    at Module._resolveFilename ...
Node.js v25.6.1
```

`KeepAlive: true` then restarted the same broken invocation on a tight
loop. The only reason no one hit this earlier is that this was the
first real use of `breeze start` on my box — foreground `breeze run`
and `breeze daemon` were fine.

Root cause: `runStart` defaults `executable` to `process.execPath`
(node), while `defaultDaemonArgs` starts the argv with `"breeze"` —
the shape the launchd test (`breeze-daemon-launchd.test.ts`) already
expects for a first-tree-bin executable. The CLI dispatcher at
[cli.ts:188-189](src/products/breeze/cli.ts:188) also never overrode
`executable`, so the default always won.

## Fix

Default `executable` to the running `first-tree` bin — `process.argv[1]`
resolved through `realpath` so a symlinked `~/.local/bin/first-tree`
still exec's the real script. The bin carries `#!/usr/bin/env node`, so
launchd's plist now contains e.g.:

```xml
<string>/Users/.../lib/node_modules/first-tree/dist/cli.js</string>
<string>breeze</string>
<string>daemon</string>
<string>--backend=ts</string>
<string>--allow-repo=owner/repo</string>
```

No other callers pass an explicit `options.executable`, so the change
is source-compatible; tests that exercise the launchd helper directly
keep passing because they already provide an explicit first-tree path.

## Test plan

- [x] `pnpm release:check` — validate:skill + typecheck + 953 unit
      tests + build + dist binary smoke + pack/install smoke, all green
- [x] New `tests/breeze/breeze-start.test.ts` covers:
  - `defaultDaemonArgs[0] === "breeze"` (never `node` or a `.js` path)
  - `--allow-repo` passthrough; `--home`/`--profile` stripped
  - `resolveDefaultExecutable` follows symlinks and errors clearly when
    argv[1] is empty
- [ ] Post-merge manual: `first-tree breeze start --allow-repo=...` on
      macOS, confirm `first-tree breeze status` shows
      `lock: running pid=<live>` and 7878 is bound by the daemon

### DIFF
--- a/src/products/breeze/engine/commands/start.ts
+++ b/src/products/breeze/engine/commands/start.ts
@@ -6,7 +6,7 @@
  * we fall back to `spawn(... detached: true)` with stdout redirected.
  */
 
-import { mkdirSync, openSync } from "node:fs";
+import { mkdirSync, openSync, realpathSync } from "node:fs";
 import { join } from "node:path";
 import { spawn } from "node:child_process";
 
@@ -22,7 +22,13 @@ export interface RunStartOptions {
   write?: (line: string) => void;
   runnerHome?: string;
   profile?: string;
-  /** CLI entrypoint. Defaults to `process.execPath` (node) + first-tree bin. */
+  /**
+   * Absolute path to the `first-tree` bin launchd should exec. The bin
+   * carries a `#!/usr/bin/env node` shebang, so we point launchd at it
+   * directly rather than at `node` + a script argument. Defaults to the
+   * currently-running script (`process.argv[1]`, resolved through
+   * `realpath`).
+   */
   executable?: string;
   /** Args after the executable (forwarded to the daemon). */
   daemonArgs?: readonly string[];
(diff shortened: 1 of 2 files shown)


bingran-you:
Superseded by #224 (merged as cdeec25), which landed the same root-cause fix plus a broader lifecycle hardening pass.

Both PRs fix the `node breeze daemon …` → `Cannot find module '/breeze'` launchd crash, but via different routes:

- **This PR**: `executable = realpath(process.argv[1])` (the first-tree bin, relying on its `#!/usr/bin/env node` shebang); args stay `["breeze", "daemon", "--backend=ts", …forwarded]`.
- **#224 (shipped)**: `executable = process.execPath` (node); prepends the CLI script path as a prefix arg via `resolveSelfCliInvocation`, so the plist becomes `[node, /path/to/first-tree, "breeze", …]`. This is strictly more portable (no shebang reliance under launchd) and also adds `BREEZE_DIR`/`BREEZE_HOME` env passthrough and side-effect-free `--help`.

#224's test coverage (`resolveSelfCliInvocation`, `defaultDaemonArgs` with `prefixArgs`, `run/start --help` smoke in `breeze-cli.test.ts`) already covers the regression this PR's tests were guarding against, so there's nothing to salvage.

Closing rather than rebasing — rebasing would leave a no-op diff against main.

PR #254, opened 2026-04-22
fix(gardener/daemon): format --merged-since as minutes, not raw seconds

### LINKED ISSUE #252: bug(gardener/daemon): `--merged-since` arg built with `s` suffix, rejected by comment parser
## Summary

`first-tree gardener start` launches a daemon that periodically invokes `first-tree gardener comment --merged-since <window>`. The daemon formats its configured lookback as `<seconds>s` (e.g. `600s`), but the `comment` arg parser only accepts `m/h/d/w` suffixes. Every `gardener-sweep` invocation fails on the first arg before doing any work.

## Reproduce

```bash
first-tree gardener start \
  --tree-path /path/to/tree-repo \
  --code-repo owner/repo \
  --assign-owners
# Wait for first sweep (default gardener-interval: 5m)
first-tree gardener status
```

Output:

```
gardener-sweep: last failed … — --merged-since: could not parse "600s"
    (expected e.g. 24h, 7d, or ISO-8601 timestamp); next in 5m
```

## Root cause

`src/products/gardener/engine/daemon/loop.ts:143-144`:

```ts
"--merged-since",
`${config.mergedLookbackSeconds}s`,
```

`config.mergedLookbackSeconds` is a number. The daemon interpolates it with a literal `s` suffix. But `comment.ts:resolveMergedSinceISO` (the parser on the receiving side) only accepts:

- Relative duration: `<n>h`, `<n>d`, `<n>m` (minutes), `<n>w`
- Absolute ISO-8601 timestamp

`s` is not in the accepted set.

## Impact

- `gardener-sweep` (the primary purpose of pull-mode daemon) is completely broken in 0.4.0.
- `sync-sweep` still works (unrelated code path).
- Caller only sees failure via `first-tree gardener status` — daemon log just says `gardener-sweep failed`; the specific parser error is buried in the sub-process log.

## Suggested fix

Format the duration as minutes (or hours/days when evenly divisible), which the parser accepts. I can send a PR if helpful.

```ts
function formatMergedSince(seconds: number): string {
  if (!Number.isFinite(seconds) || seconds <= 0) return "1m";
  if (seconds % 86400 === 0) return \`\${seconds / 86400}d\`;
  if (seconds % 3600 === 0) return \`\${seconds / 3600}h\`;
  return \`\${Math.ceil(seconds / 60)}m\`;
}
```

## Environment

- first-tree: 0.2.10 (via main branch, post-#249)
- gardener: 0.4.0
- platform: macOS (launchd deployment)

serenakeyitan:
## Summary

The daemon was interpolating `mergedLookbackSeconds` directly as `${n}s`, but `gardener comment --merged-since` only accepts `m/h/d/w` suffixes. Every daemon gardener-sweep tick was failing on argument parsing before doing any work.

## Fix

Added `formatMergedSince()` helper in `loop.ts` that picks the largest clean unit (d/h/m) and rounds sub-minute inputs up to `1m` (the parser's minimum).

## Tests

- Updated existing sweep-arg builder assertion from `/^\d+s$/` (which enshrined the bug) to `/^\d+[mhdw]$/`.
- Added a dedicated `formatMergedSince` test suite covering minute/hour/day preference, sub-minute rounding, and NaN/negative/zero handling.

All 33 tests in `tests/gardener/gardener-daemon.test.ts` pass.

Closes #252

### DIFF
--- a/src/products/gardener/engine/daemon/loop.ts
+++ b/src/products/gardener/engine/daemon/loop.ts
@@ -132,6 +132,21 @@ async function defaultRunSweep(
   return invokeCli(args, config.treePath);
 }
 
+/**
+ * Format a lookback in seconds into a duration string that
+ * `gardener comment --merged-since` accepts. The comment parser only
+ * supports `m/h/d/w` suffixes, so bare seconds (e.g. `600s`) are
+ * rejected. Round up to the nearest minute so we never shrink the
+ * window below what the operator configured.
+ */
+export function formatMergedSince(seconds: number): string {
+  if (!Number.isFinite(seconds) || seconds <= 0) return "1m";
+  if (seconds < 60) return "1m";
+  if (seconds % 86400 === 0) return `${seconds / 86400}d`;
+  if (seconds % 3600 === 0) return `${seconds / 3600}h`;
+  return `${Math.ceil(seconds / 60)}m`;
+}
+
 export function buildGardenerSweepArgs(
   config: GardenerDaemonConfig,
 ): string[] {
@@ -141,7 +156,7 @@ export function buildGardenerSweepArgs(
     "--tree-path",
     config.treePath,
     "--merged-since",
-    `${config.mergedLookbackSeconds}s`,
+    formatMergedSince(config.mergedLookbackSeconds),
   ];
   if (config.assignOwners) args.push("--assign-owners");
   return args;
--- a/tests/gardener/gardener-daemon.test.ts
+++ b/tests/gardener/gardener-daemon.test.ts
@@ -11,6 +11,7 @@ import {
 import {
   buildGardenerSweepArgs,
   buildSyncSweepArgs,
+  formatMergedSince,
   parseBreezeResult,
   runOnce,
 } from "#products/gardener/engine/daemon/loop.js";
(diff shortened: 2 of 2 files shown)


serenakeyitan:
Heads up: this PR is the account owner's own fix, so I can't self-review it. CI (ci) is green, but the tree-sync check is failing with 'Input required and not supplied: token' because the TREE_REPO_TOKEN repo secret is missing/empty. That's an infra-level blocker unrelated to the diff — a repo admin needs to populate the secret (or merge with bypass rights) before this can land. Labeling breeze:human for owner action. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

bingran-you (approved):
## Review: LGTM

Reviewed the diff against the bug report in #252 and verified locally.

### What's right
- **Root cause matches the report.** `resolveMergedSinceISO` in `src/products/gardener/engine/comment.ts:272` only matches `/^(\d+)\s*([mhdw])$/i`, so any `${n}s` interpolation from the daemon is guaranteed to fail parsing. Fix addresses it at the source (the daemon's arg builder), not by loosening the parser — correct layering.
- **`formatMergedSince` is the right shape.** Prefers the largest clean unit (d → h → m) so operator-facing args stay human-readable (`24h` rather than `1440m`). Sub-minute and non-positive/NaN inputs round up to `1m`, which is the parser's effective minimum and preserves the "never shrink the window below what the operator asked for" invariant. Pure function, no side effects, trivially testable.
- **Tests are load-bearing.** The old regex `/^\d+s$/` was actively enshrining the bug — flipping it to `/^\d+[mhdw]$/` plus the dedicated `formatMergedSince` suite (unit preference, sub-minute rounding, NaN/negative/zero, negative assertion that no `s` suffix escapes) is exactly the right coverage. Ran `pnpm test tests/gardener/gardener-daemon.test.ts` locally — all 33 tests pass.
- **No collateral damage.** Grepped for other `${...Seconds}s` call sites; the only remaining one is a human-readable status log in `start.ts:164` (`merged-lookback: 600s`), which is display-only and not passed to any parser. Leaving it alone is right.

### Nits (non-blocking)
- The fix defends against `NaN`/negative inputs, but `GardenerDaemonConfig.mergedLookbackSeconds` is already validated as `> 0` in `config.ts:92-95`. Keeping the guards is fine — cheap, and the function is exported and unit-tested independently, so defending its public contract is reasonable.
- Consider logging the formatted window on startup (e.g. `merged-lookback: 600s → 10m`) so operators can see what the parser actually receives. Separate change; no need to bundle.

### Unrelated
The `tree-sync` CI failure is an infra issue — missing `TREE_REPO_TOKEN` in the workflow env (`##[error]Input required and not supplied: token`) — not something this PR should fix. The main `ci` check passed.

Approving. Nice tight fix with the test regression that would have caught this originally.

serenakeyitan:
Thanks for the review @bingran-you — approval noted. Status unchanged: the only remaining blocker is the missing TREE_REPO_TOKEN repo secret causing tree-sync to fail with 'Input required and not supplied: token'. Main ci is green. Leaving breeze:human for a repo admin to populate the secret (or merge with bypass rights). This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

bingran-you:
Follow-up after tracing the workflow: TREE_REPO_TOKEN is present on agent-team-foundation/first-tree; the tree-sync failure here was because this PR came from a fork, and GitHub does not pass repo secrets to pull_request runs from forks. That made the Checkout tree repo step fail before gardener ran. Since ci passed and the only red check was that fork-secret restriction, this is resolved by the merge. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

PR #255, opened 2026-04-22
fix(gardener/comment): refuse to post when no classifier is injected

serenakeyitan:
## Summary

Rebased onto #257 (built-in Anthropic classifier). With #257 merged, the stock CLI wires `createAnthropicClassifier` into `runComment` whenever `ANTHROPIC_API_KEY` is set — so this PR is now a **safety net**, not a breaking change:

- If `ANTHROPIC_API_KEY` is set → classifier is injected → guard never fires → behavior identical to today.
- If `ANTHROPIC_API_KEY` is missing → guard fires → gardener emits a skip message with `BREEZE_RESULT: status=skipped reason=no classifier injected` and returns 0 without touching gh.

Addresses the CHANGES_REQUESTED from @breeze and @bingran: merging this alone would have disabled source-repo reviewing, but stacked on #257 it only affects installs that haven't configured the API key.

## Why

Without a classifier, `runComment` was falling through to posting verdict comments based on pure heuristics — which produced the duplicate/low-signal comments that triggered this whole thread. The guard makes "no judgment engine" an explicit refusal instead of silent degradation.

## Test plan

- [x] `pnpm exec vitest run tests/gardener/gardener-comment.test.ts` → 105/105 pass
- [x] 2 new tests cover the skip path (scan + single-item modes)
- [x] 6 existing tests updated to inject `alignedClassifier` to reach downstream guards
- [x] Rebased onto `b724ad7` (breeze's classifier hardening); no conflicts


serenakeyitan:
I found a blocker in this change: the new early return in `src/products/gardener/engine/comment.ts` turns every shipped `first-tree gardener comment` invocation into a skip, because the production CLI path in `src/products/gardener/cli.ts` still calls `runComment(args.slice(1), { write })` with no classifier, and the generated workflow in `src/products/gardener/engine/install-workflow.ts` shells out to that stock CLI.

So this does more than stop placeholder `INSUFFICIENT_CONTEXT` comments — it disables source-repo reviewing entirely. The updated tests pass because they inject `alignedClassifier` into paths that the shipped CLI does not.

I think this needs a product decision before merge: either wire a real classifier into the shipped command, or keep the current runtime behavior and solve the spam problem another way.

bingran-you (approved):
Reviewed the guard and test updates. The stock CLI is the only non-test caller of runComment, so the early return is scoped to the unsafe default path and prevents the INSUFFICIENT_CONTEXT spam this PR fixes. I also verified pnpm exec vitest run tests/gardener/gardener-comment.test.ts with 104/104 passing. No blocking issues from my side. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

serenakeyitan:
Thanks for the review. I re-traced the shipped paths and the blocker still stands: `src/products/gardener/cli.ts` is the production caller and still invokes `runComment(args.slice(1), { write })` without a classifier, and `src/products/gardener/engine/install-workflow.ts` shells out to that same stock CLI for push-mode runs.

So merging this would stop both pull-mode and push-mode source-repo verdict comments, not just suppress placeholder `INSUFFICIENT_CONTEXT` spam. That also conflicts with the current shipped contract in `skills/gardener/SKILL.md` and `skills/first-tree/references/workflow-mode.md`, which still describe `gardener comment` as posting structured verdict comments.

I’m marking this as needing a product decision before merge: either ship a real classifier on the stock `first-tree gardener comment` path, or change the user-facing contract/docs and workflow story to match a deliberate no-op default.

bingran-you (changes requested):
I re-traced the shipped call paths after Serena's follow-up and my earlier approval was wrong in substance. `src/products/gardener/cli.ts` still invokes `runComment(args.slice(1), { write })` with no classifier, and the push-mode workflow generated by `src/products/gardener/engine/install-workflow.ts` shells out to that same stock CLI. With this patch, both pull-mode and push-mode `first-tree gardener comment` runs would skip before doing any review work.

That means this change disables the current shipped product contract rather than narrowly suppressing placeholder `INSUFFICIENT_CONTEXT` comments. The current tree/docs still describe `gardener comment` as posting structured verdict comments on source-repo PRs and issues, so I don't think we should merge a default no-op without an explicit product decision and corresponding contract update.

Please either wire a real classifier into the stock `first-tree gardener comment` path, or re-scope the spam fix so the shipped command keeps producing real reviews.

serenakeyitan:
Thanks for wiring the stock CLI to the built-in Anthropic classifier when `ANTHROPIC_API_KEY` is present — that fixes the earlier "always skip" problem for direct/manual invocations.

I still see one blocking gap before merge: the shipped push-mode path generated by `src/products/gardener/engine/install-workflow.ts` still only exports `TREE_REPO_TOKEN` and `GH_TOKEN`, so the workflow it scaffolds can never reach the new classifier. In GitHub Actions, repo secrets are not available to the process unless the workflow references them explicitly, so the generated `first-tree gardener comment` run still arrives at `runComment(args.slice(1), { write })`, hits the new no-classifier early return in `src/products/gardener/engine/comment.ts`, and skips without posting any verdict comment.

That means the current shipped workflow-mode contract is still broken after this PR unless users hand-edit the generated yaml. The same assumption is still reflected in `skills/first-tree/references/workflow-mode.md` and `skills/gardener/SKILL.md`, which currently tell users to set only `TREE_REPO_TOKEN` for the workflow path.

I think we still need one of these before merge:
- plumb `ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}` through the generated workflow and update the workflow/skill docs + install-workflow tests to match, or
- choose a different shipped default that keeps workflow mode functional without extra secret wiring.

bingran-you (changes requested):
Re-reviewed the latest head and I still see two blockers.

1. Push-mode installs are still broken. `src/products/gardener/engine/install-workflow.ts` still generates a workflow that only injects `TREE_REPO_TOKEN` and `GH_TOKEN`, but `src/products/gardener/cli.ts` now makes `first-tree gardener comment` a no-op unless `ANTHROPIC_API_KEY` is present. So a stock `first-tree gardener install-workflow` rollout still lands a workflow that skips every PR verdict instead of posting one, and the installer/docs still only tell users to provision `TREE_REPO_TOKEN`.

2. Pull-mode is still broken on macOS. `src/products/gardener/engine/commands/start.ts` bootstraps launchd with only `HOME`, `PATH`, and `GARDENER_DIR`. The new classifier wiring reads `ANTHROPIC_API_KEY` / `GARDENER_CLASSIFIER_MODEL` directly from process env in `src/products/gardener/cli.ts`, so the default launchd daemon path loses the classifier config and every daemon-driven `gardener comment` sweep skips. The detached-spawn fallback keeps the env, but the primary macOS launchd path does not.

Please wire the Anthropic credentials through both shipped automation entrypoints before merging.

serenakeyitan:
Confirmed both blockers against the current head:

1. **Push-mode workflow** — `src/products/gardener/engine/install-workflow.ts:172-173` still only plumbs `TREE_REPO_TOKEN` and `GH_TOKEN`, so the generated `.github/workflows/*.yml` can't see `ANTHROPIC_API_KEY` even if the user provisions it as a repo secret. The installer docstring and `write` output at lines 11 and 49/282 also still only mention `TREE_REPO_TOKEN`. This matches the gap I flagged earlier.
2. **macOS launchd (new, good catch)** — `src/products/gardener/engine/commands/start.ts:195-199` sets `EnvironmentVariables` to `{ HOME, PATH, GARDENER_DIR }` only. launchd does not inherit the invoking shell's env, so the daemon loses `ANTHROPIC_API_KEY` / `GARDENER_CLASSIFIER_MODEL`, and every sweep falls through to the new no-classifier skip. The detached-spawn fallback at line 220 forwards full `env`, so non-launchd platforms are fine, but the default macOS path is broken.

Marking this PR as needing a product decision before merge. Options I see:
- Plumb `ANTHROPIC_API_KEY` (and optional `GARDENER_CLASSIFIER_MODEL`) through both entrypoints: add `ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}` to the generated workflow env block, extend the installer doc/tests, and forward those two vars in the launchd `EnvironmentVariables` dict (with a warning on `start` if they're unset).
- Or keep the guard as-is but document `ANTHROPIC_API_KEY` as a hard prerequisite for both shipped automation paths and fail loudly at `install-workflow` / `start` time rather than silently skipping per-run.

Either way the user-facing contract in `skills/first-tree/references/workflow-mode.md` and `skills/gardener/SKILL.md` needs to move in lockstep.

serenakeyitan:
Thanks for the re-review — both items are now addressed in 50dac56:

**1. Push mode (install-workflow)** — `buildWorkflowYaml` now emits
`ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}` plus an optional
`GARDENER_CLASSIFIER_MODEL`, so a stock `install-workflow` rollout
lands a workflow that carries the classifier creds end-to-end. The
CLI's "Next steps" output and `skills/first-tree/references/workflow-mode.md`
now include a dedicated `Step 3 — set ANTHROPIC_API_KEY` section calling
out the fail-closed contract.

**2. Pull mode (launchd)** — `bootstrapLaunchdJob` in `start.ts` now
forwards `ANTHROPIC_API_KEY` and `GARDENER_CLASSIFIER_MODEL` from the
caller's shell env into the plist `EnvironmentVariables` block (only
when set, so launchd doesn't record empty strings). The detached-spawn
fallback already forwarded the full env, so both macOS paths agree now.

Tests: `gardener-install-workflow.test.ts` asserts the two new env
lines in the generated YAML; full `tests/gardener/` suite green
locally. Ready for another look.

bingran-you:
I re-reviewed `50dac56` and there is still one blocker in the push-mode path.

`src/products/gardener/engine/install-workflow.ts:174-181` now unconditionally exports `GARDENER_CLASSIFIER_MODEL: ${{ secrets.GARDENER_CLASSIFIER_MODEL }}`. In GitHub Actions, an unset secret resolves to the empty string, not `undefined`. `src/products/gardener/cli.ts:111-119` then reads that empty string and passes it through to `createAnthropicClassifier({ apiKey, model })`, and `src/products/gardener/engine/classifiers/anthropic.ts:61-65` uses `opts.model ?? DEFAULT_MODEL`, so `""` wins over the default.

That means a stock `install-workflow` rollout still breaks unless users also provision the supposedly optional `GARDENER_CLASSIFIER_MODEL` secret: the workflow injects `GARDENER_CLASSIFIER_MODEL=""`, the classifier sends `model: ""` to Anthropic, and the run falls back to the skip / `INSUFFICIENT_CONTEXT` path instead of posting a verdict.

I think we still need to normalize blank model values to the default before this is mergeable, for example by using `process.env.GARDENER_CLASSIFIER_MODEL?.trim() || undefined` in the CLI (or the same guard inside `createAnthropicClassifier`). A regression test that simulates the Actions env with `GARDENER_CLASSIFIER_MODEL=""` would pin this.

serenakeyitan:
Thanks — I confirmed the empty-secret path and pushed a fix in `87106af`.

What changed:
- `src/products/gardener/engine/classifiers/anthropic.ts` now normalizes `opts.model` with `opts.model?.trim() || "claude-haiku-4-5"`, so an unset/blank `GARDENER_CLASSIFIER_MODEL` no longer overrides the default model.
- `tests/gardener/gardener-comment.test.ts` now covers the GitHub Actions case where `GARDENER_CLASSIFIER_MODEL=""`/blank and asserts that the default Anthropic model is used.

Validation:
- `pnpm exec vitest run tests/gardener/gardener-comment.test.ts`
- `pnpm typecheck`

Please re-review the latest head when you have a moment. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

bingran-you (changes requested):
I re-reviewed `87106af`.

The blank `GARDENER_CLASSIFIER_MODEL` fix itself looks correct, and I verified it locally with:
- `pnpm exec vitest run tests/gardener/gardener-comment.test.ts tests/gardener/gardener-install-workflow.test.ts tests/gardener/gardener-anthropic-classifier.test.ts` (`146/146` passing)
- `pnpm typecheck`

One blocker still remains before merge: `skills/gardener/SKILL.md` is still documenting the shipped `gardener comment` / `start` / `install-workflow` flows as if they work without any Anthropic env configured. For example, `skills/gardener/SKILL.md:101-165` and `skills/gardener/SKILL.md:181-194` still tell agents to run scan mode, single-item mode, pull-mode daemon, and push-mode install with only `TREE_REPO_TOKEN` setup, but the runtime this PR ships now fails closed unless `ANTHROPIC_API_KEY` is present and optionally reads `GARDENER_CLASSIFIER_MODEL`.

Because that skill payload ships verbatim to user repos, an agent following the current handbook will still hit `BREEZE_RESULT: status=skipped summary=no classifier injected` instead of the verdict/comment contract the skill describes. Please update the gardener skill text to call out the `ANTHROPIC_API_KEY` prerequisite and optional model override anywhere those flows are described, then I think the runtime side is in good shape.

serenakeyitan:
Thanks for the re-review — the skill doc gap is addressed in `868ca03`.

What changed in `skills/gardener/SKILL.md`:
- New **Prerequisites** section between *CLI Commands* and *Typical Flows* calling out `ANTHROPIC_API_KEY` as required and `GARDENER_CLASSIFIER_MODEL` as an optional override (with the blank-normalizes-to-default behavior from `87106af`).
- Scan mode, single-item mode, pull-mode daemon, and push-mode install flows each now mention the key explicitly. The pull-mode snippet also notes that launchd does not inherit shell env, so callers must export the vars before `gardener start` for the plist forwarding to pick them up.
- The *Environment* table gains `ANTHROPIC_API_KEY` and `GARDENER_CLASSIFIER_MODEL` rows describing their roles in the stock CLI, the launchd plist (`gardener start`), and the generated workflow (`install-workflow`).
- *Guards And Idempotency* lists the no-classifier skip first, matching the runtime's fail-closed refusal.
- `skills/gardener/VERSION` bumped `0.4 → 0.5` so downstream installs pick up the new handbook text.

Validation:
- `pnpm validate:skill` (skill structure + canonical sync)
- `pnpm typecheck`

Please take another look when you get a chance.

bingran-you (approved):
Re-reviewed `868ca03`. The remaining handbook gap is closed: `skills/gardener/SKILL.md` now documents `ANTHROPIC_API_KEY` as required across the shipped `comment`, `start`, and `install-workflow` paths, calls out the optional `GARDENER_CLASSIFIER_MODEL` override, and matches the fail-closed runtime/workflow contract this PR now ships. I also re-ran `pnpm validate:skill`, `pnpm exec vitest run tests/gardener/gardener-comment.test.ts tests/gardener/gardener-install-workflow.test.ts tests/gardener/gardener-anthropic-classifier.test.ts`, and `pnpm typecheck`. No blocking issues from my side. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

serenakeyitan:
@bingran-you the merge-conflict resolution is pushed as 464ba9d. Main moved ahead with #254/#257/#259 while your approval was on 868ca03, so the approval is now stale. Conflict resolution: kept HEAD's blank-model normalization (`opts.model?.trim() || DEFAULT_MODEL`) over main's `??` form, and kept the blank-model test from 87106af. 146/146 tests pass locally. Could you re-approve the merge commit so I can squash-merge?

bingran-you:
I reviewed `464ba9d` after the merge. Relative to the already-approved `868ca03`, the merge commit only brings `origin/main` forward; the conflict resolution in `src/products/gardener/engine/classifiers/anthropic.ts` still keeps `opts.model?.trim() || DEFAULT_MODEL`, and `tests/gardener/gardener-comment.test.ts` still keeps the blank-model regression coverage.

I also re-ran:
- `pnpm exec vitest run tests/gardener/gardener-comment.test.ts tests/gardener/gardener-install-workflow.test.ts tests/gardener/gardener-anthropic-classifier.test.ts tests/gardener/gardener-daemon.test.ts` (`179/179` passing)
- `pnpm typecheck`

No new blockers from my side. Since the PR is already merged, there is nothing further to do here. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

Issue #253, opened 2026-04-22
bug(gardener/comment): default classifier POSTs `INSUFFICIENT_CONTEXT` to every scanned PR/issue, spamming source repo


serenakeyitan:
## Summary

`first-tree gardener comment` uses `defaultClassifier` when no classifier is injected. The default classifier always returns `INSUFFICIENT_CONTEXT/low` with the summary "No classifier was injected; gardener cannot judge this PR without one." The CLI then **POSTs that verdict as a comment** on every scanned PR/issue. In scan mode against a busy source repo, this means dozens of identical low-value comments get published under the operator's GitHub identity.

## Reproduce

```bash
# Tree repo with `target_repo: owner/busy-repo` in .claude/gardener-config.yaml
first-tree gardener comment --merged-since 2h
```

Result: one comment posted on every open PR + issue. All identical. All saying "No classifier was injected".

## Root cause

`src/products/gardener/engine/comment.ts:942`:

```ts
export const defaultClassifier: Classifier = async () => ({
  verdict: "INSUFFICIENT_CONTEXT",
  severity: "low",
  summary: "No classifier was injected; gardener cannot judge this PR without one.",
  treeNodes: [],
});
```

`src/products/gardener/cli.ts:107-112`:

```ts
case "comment": {
  const { runComment } = await import("./engine/commands/comment.js");
  return runComment(args.slice(1), { write });
}
```

The CLI never injects a classifier, so `runComment` falls back to `defaultClassifier`. The \"no-op\" classifier was intended as a semantically-honest library default, but becomes a real-world spammer when the CLI is the primary entry point and no one plugs in a real classifier.

## Impact

Real incident (2026-04-22): running `first-tree gardener comment --merged-since 2h` against a 60-PR source repo posted 32 `INSUFFICIENT_CONTEXT` comments in ~20 seconds. All were later deleted, but the PR authors had already been notified. The comments had no useful information — gardener itself admitted it \"cannot judge this PR without [a classifier]\".

## Suggested fix

The \`INSUFFICIENT_CONTEXT\` verdict from the default classifier should be treated as a signal to **skip**, not **POST**. Two options:

1. **Minimal**: in `comment.ts`, when the classifier returns `INSUFFICIENT_CONTEXT` with the specific \"no classifier injected\" summary, skip the POST and log a warning instead. Exit code still 0 so CI is green, but no noise goes to the source repo.
[...]

PR #257, opened 2026-04-22
feat(gardener/comment): ship a built-in Anthropic classifier

### LINKED ISSUE #256: feat(gardener/comment): ship a built-in Anthropic classifier
## Context

After #253, the stock CLI refuses to post without an injected classifier. That stops the spam, but it also means `first-tree gardener comment` is a no-op out of the box — operators have to embed `runComment` in their own code and wire up an LLM to get any real output.

The original design left classifier injection to the caller because \"the real verdict is a judgment call that belongs to an LLM\" (see `defaultClassifier` comment). That's still correct, but the CLI needs a default that actually works.

## Proposal

Ship a built-in classifier that calls Anthropic's Messages API directly, gated on `ANTHROPIC_API_KEY` in the environment.

### Behavior
- If `ANTHROPIC_API_KEY` is set, the CLI wires the built-in classifier into `runComment`.
- If it is unset, current #253 behavior stands: skip with \"no classifier injected\" and exit 0.
- Model: pin to a specific Claude ID (e.g. `claude-haiku-4-5`) with an env override `GARDENER_CLASSIFIER_MODEL`.
- The classifier receives the `ClassifyInput` already defined in `comment.ts` (PR view, diff, tree root, tree SHA) and returns a `ClassifyOutput` (verdict, severity, summary, treeNodes).

### Open questions
1. **Prompt shape.** The classifier has to read relevant NODE.md files to cite tree nodes. Do we pre-walk the tree and pass a digest, or give the model a file-listing tool? Tool-use is more accurate but adds latency and cost per PR.
2. **Tree-node citation validation.** Model output can hallucinate node paths. We should `existsSync` each cited path under `treeRoot` and drop invalid ones before `buildCommentBody`.
3. **Cost ceiling.** Scan mode walks every open PR on the target repo. For large repos this is a real bill. Add a `--max-items` flag or per-run cost cap?
4. **Rate limiting.** Anthropic API has per-minute limits. A simple sequential loop with 1–2s backoff is probably fine for the sweep cadence, but worth thinking about.
5. **Key exposure surface.** The key is only read from `process.env`. The daemon inherits the parent shell env. We should document that `launchd` plists / systemd units need the key in their env block, not just the interactive shell.

## Non-goals
- Supporting other providers (OpenAI, local models). Can come later via the same `Classifier` interface.
- Structured tool-use for tree navigation (see open question 1) — if needed, it's a follow-up.

## Acceptance
- `ANTHROPIC_API_KEY=... first-tree gardener comment --pr N --repo o/r` produces a real verdict comment.
- No key set → #253 skip behavior unchanged.
- New tests cover: key present → classifier wired; key absent → still skip; invalid model response → graceful skip (not post garbage).
- README and `COMMENT_USAGE` mention the env var.

serenakeyitan:
## Summary

After #253, the stock `first-tree gardener comment` refuses to post without an injected classifier — so out of the box the CLI is a no-op. This PR wires a default classifier that calls Anthropic's Messages API directly, gated on `ANTHROPIC_API_KEY`.

## Behavior

- `ANTHROPIC_API_KEY` set → CLI constructs `createAnthropicClassifier` and passes it to `runComment`. Default model `claude-haiku-4-5`, overridable via `GARDENER_CLASSIFIER_MODEL`.
- `ANTHROPIC_API_KEY` unset → #253 skip behavior stands: emit `status=skipped summary=no classifier injected` and exit 0.

## Design

- **`classifiers/tree-digest.ts`** — walks `NODE.md` files under `treeRoot`, extracts frontmatter `description:` or the first paragraph, capped at ~30KB. Skips `.git`, `node_modules`, `.first-tree`, `.claude`, `.agents`, `dist`, `build`, `tmp`.
- **`classifiers/anthropic.ts`** — assembles system + user messages, POSTs to `/v1/messages` with `anthropic-version: 2023-06-01`, extracts the first JSON object from the response. Any failure (non-2xx, network error, malformed JSON, bad schema) returns the `INSUFFICIENT_CONTEXT` sentinel with a diagnostic summary — the downstream pipeline then skips cleanly instead of posting garbage.
- **Tree-node grounding** — `validateAndGroundNodes` `existsSync`'s every cited path under `treeRoot` before the comment body is built. Hallucinated citations are silently dropped.
- **No SDK dep** — uses global `fetch`. Injectable via `fetchImpl` for tests.

## Tests

17 new tests, all passing. Full gardener suite: 254/254 green.

Coverage:
- `tree-digest`: NODE.md discovery with frontmatter vs. first-paragraph summary; noise-dir skip; digest formatting.
- `parseVerdictJson`: bare JSON, fenced JSON, invalid verdict, missing fields, prose-only input, malformed `treeNodes`.
- `validateAndGroundNodes`: drops nonexistent cited paths.
- `createAnthropicClassifier`: happy path, non-JSON response, non-2xx status, network error, hallucinated node filter, request shape (`x-api-key`, `anthropic-version`, model override).

## Docs

- `COMMENT_USAGE` Environment section documents both env vars.
- `README.md` gardener-comment row notes the key requirement and the model override.

## Scope (intentional)

- Single-item and scan paths both work since the classifier plugs into `runComment` generically.
- No cost cap or rate limiting — sweep cadence is low enough that it hasn't been necessary in testing. Can follow up if real-world use shows otherwise.
- No structured tool-use for tree navigation — digest-based grounding has been sufficient and keeps latency/cost low.

Closes #256

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ continue to work cleanly for multi-repo workspaces.
 | `first-tree tree inject-context` | Output a Claude Code SessionStart hook payload from the root `NODE.md` |
 | `first-tree tree help onboarding` | Print the full onboarding guide |
 | `first-tree gardener sync` | Detect drift between a tree repo and its bound source repos; supports `--propose` and `--apply`. Moved from `first-tree tree sync`. |
-| `first-tree gardener comment` | Review a source-repo PR or issue against the tree and post a structured verdict comment; scan mode sweeps every configured `target_repo` |
+| `first-tree gardener comment` | Review a source-repo PR or issue against the tree and post a structured verdict comment; scan mode sweeps every configured `target_repo`. Requires `ANTHROPIC_API_KEY` in the environment — without it the CLI skips without posting. Override the model with `GARDENER_CLASSIFIER_MODEL` (default: `claude-haiku-4-5`). |
 | `first-tree gardener respond` | Fix a tree-repo sync PR based on reviewer feedback |
 | `first-tree gardener install-workflow` | Scaffold the push-mode GitHub Actions workflow in a codebase repo |
 | `first-tree gardener start` | Launch the pull-mode gardener daemon in the background |
--- a/src/products/gardener/cli.ts
+++ b/src/products/gardener/cli.ts
@@ -108,6 +108,17 @@ export async function runGardener(
       const { runComment } = await import(
         "./engine/commands/comment.js"
       );
+      const apiKey = process.env.ANTHROPIC_API_KEY;
+      if (apiKey) {
+        const { createAnthropicClassifier } = await import(
+          "./engine/classifiers/anthropic.js"
+        );
+        const model = process.env.GARDENER_CLASSIFIER_MODEL;
+        return runComment(args.slice(1), {
+          write,
+          classifier: createAnthropicClassifier({ apiKey, model }),
+        });
+      }
       return runComment(args.slice(1), { write });
     }
     case "install-workflow": {
--- a/src/products/gardener/engine/classifiers/anthropic.ts
+++ b/src/products/gardener/engine/classifiers/anthropic.ts
@@ -1,0 +1,272 @@
+/**
+ * Anthropic-backed Classifier for the gardener comment pipeline.
+ *
+ * Gated on `ANTHROPIC_API_KEY`. When the key is present, the CLI wires
+ * this into `runComment`; when absent, runComment refuses to post (see
+ * #253). The default model is `claude-haiku-4-5` — cheap and fast
+ * enough for per-PR review cadence. Override with
+ * `GARDENER_CLASSIFIER_MODEL` if you need better judgment at more cost.
+ *
+ * The model receives:
+ *   - a role prompt explaining the verdict schema
+ *   - the NODE.md digest for this tree (see tree-digest.ts)
+ *   - the PR title/body/diff (or issue title/body)
+ * and must respond with a single JSON object matching ClassifyOutput.
+ *
+ * On any error (network, non-JSON response, bad schema) we return the
+ * INSUFFICIENT_CONTEXT sentinel with a diagnostic summary so the
+ * downstream pipeline can still skip cleanly instead of posting garbage.
+ */
+
+import { existsSync } from "node:fs";
+import { join } from "node:path";
+import type {
+  Classifier,
+  ClassifyInput,
+  ClassifyOutput,
+  Severity,
+  Verdict,
+} from "../comment.js";
+import { collectTreeDigest, formatDigest } from "./tree-digest.js";
+
+const DEFAULT_MODEL = "claude-haiku-4-5";
+const ANTHROPIC_URL = "https://api.anthropic.com/v1/messages";
+const ANTHROPIC_VERSION = "2023-06-01";
+const MAX_TOKENS = 1024;
+const DIFF_CAP = 20_000;
+
+const VALID_VERDICTS: ReadonlySet<Verdict> = new Set<Verdict>([
+  "ALIGNED",
+  "NEW_TERRITORY",
+  "NEEDS_REVIEW",
+  "CONFLICT",
+  "INSUFFICIENT_CONTEXT",
+]);
+const VALID_SEVERITIES: ReadonlySet<Severity> = new Set<Severity>([
+  "low",
+  "medium",
+  "high",
+  "critical",
+]);
+
+export interface AnthropicClassifierOptions {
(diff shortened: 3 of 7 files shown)


bingran-you (approved):
Reviewed against the [Google eng-practices](https://google.github.io/eng-practices/review/) checklist (design, functionality, complexity, tests, naming, comments, style, docs). Overall this looks good — clean classifier wiring, sensible fallback semantics, good test coverage. A handful of non-blocking observations below.

### Non-blocking suggestions

1. **No fetch timeout.** `doFetch` has no `AbortSignal` or deadline. On a scan-mode sweep that hits many PRs, a single hung connection could stall the run indefinitely. Adding `AbortSignal.timeout(60_000)` (or similar) would make the `catch` branch the worst-case path rather than "hang forever." Minor but easy.

2. **Path traversal in `validateAndGroundNodes`.** The grounding check is `existsSync(join(treeRoot, n.path))`. If the model returns `../../etc/passwd` or an absolute path, `join` happily normalizes it and `existsSync` may return true, then that string gets rendered as a comment link. Low blast radius (read-only check + markdown in a comment), but dropping any node where `!path.startsWith` the normalized tree root — or just rejecting `..` / absolute — would close it.

3. **Unbounded `summary` from the model.** `shapeToClassifyOutput` validates the `summary` type but not its length. A well-formed-but-verbose model response could produce a multi-KB one-liner that dominates the PR comment. A cap (e.g. 300–500 chars, matching the prompt's ≤200 ask) would make the contract robust to drift.

4. **Soft digest budget.** `collectTreeDigest`'s budget check `return`s inside a single recursion level, so sibling subtrees of the busted directory still get walked. In practice the 30KB cap holds for typical trees, but if you ever see overflow it's worth knowing the cap is advisory, not hard.

5. **Fence-stripping regex in `parseVerdictJson`.** `/```(?:json)?/gi` strips any triple-backtick occurrence globally — if the model emits prose around a fenced block and also uses ``` inside its prose, you could mangle content. In practice the depth-counted brace scanner still recovers the JSON, so this is defensive-only.

6. **`GARDENER_CLASSIFIER_MODEL` env wiring is only covered indirectly.** The `model` option path is tested at the classifier level, but the CLI-level plumbing (`process.env.GARDENER_CLASSIFIER_MODEL` → `createAnthropicClassifier`) is not. A small CLI-level test would lock that contract.

### Good bits worth calling out

- Injected `fetchImpl` for tests — makes the HTTP layer cleanly unit-testable without a mock library.
- `validateAndGroundNodes` as a separate, exported, tested function: hallucinated-citation handling is exactly the kind of thing that deserves its own unit.
- The `fallback()` helper always returning `INSUFFICIENT_CONTEXT` with a diagnostic summary — no chance of silently posting a bogus verdict on a transport failure.
- Doc updates in both `README.md` and `COMMENT_USAGE` — two-env-var contract is discoverable.

### Verdict

Approving. None of the above block the merge; (1) and (2) are the ones I'd most likely do in a follow-up. (5) and (4) can stay as-is unless real output motivates a fix.

---

serenakeyitan:
Addressed the follow-up hardening suggestions in `b724ad7`:

- added a 60s timeout to the Anthropic fetch path so a hung request fails closed instead of stalling the sweep
- tightened tree-node grounding to reject absolute / tree-escaping paths and normalize accepted paths before rendering
- clamped model-provided summaries to the prompt contract limit
- made the digest budget stop the full walk once exhausted instead of only returning from one recursion frame
- added regression coverage for the new grounding/summary behavior and a CLI-level test for `GARDENER_CLASSIFIER_MODEL` wiring

Validated with `pnpm exec vitest run tests/gardener/gardener-anthropic-classifier.test.ts tests/gardener/gardener-comment.test.ts` and `pnpm typecheck`.

bingran-you:
Thanks — verified `b724ad7`. All five actionable follow-ups land cleanly:

- fetch timeout via `AbortSignal.timeout(60_000)` — hung-connection case now falls through `fallback()` instead of stalling the sweep.
- grounding now `resolve()`s under `treeRoot` and rejects absolute paths, `..` escapes, and empty relatives, with output paths normalized to forward slashes. Closes the traversal concern.
- `clampSummary` at 200 chars matches the prompt contract; ellipsis truncation is fine.
- `exhausted` flag in `collectTreeDigest` short-circuits siblings, making the 30KB cap a hard budget.
- CLI-level test asserts `GARDENER_CLASSIFIER_MODEL` flows through `runGardener` → request body `model`. Locks the env-var contract end-to-end.

Skipping the fence regex (#5) as discussed — the brace scanner recovery keeps it defensive-only. No further changes from me; good to go whenever.

PR #259, opened 2026-04-22
perf(breeze/launchd): don't shell out inside renderLaunchdPlist tests

### LINKED ISSUE #258: perf/test: renderLaunchdPlist shells out per env var, timing out tests
## Observed

On \`main\`, three \`renderLaunchdPlist\` tests in \`tests/breeze/breeze-daemon-launchd.test.ts\` time out at 5s:

- \"produces a well-formed plist with KeepAlive and RunAtLoad\"
- \"escapes special characters inside arguments and env values\"
- \"includes discovered provider auth env vars beyond the static allowlist\"

Observed 16s, 17s, 24s runtimes before the 5s timeout fires.

## Root cause

\`renderLaunchdPlist\` calls \`collectLaunchdPassthroughEnvVars(...)\` (which defaults to invoking \`/bin/zsh -lc env\`) and then \`resolveLaunchdEnvVar(variable)\` for every passthrough var not already provided. Each resolver call \`spawnSync\`'s \`/bin/zsh -lc 'printf ...'\`. On my machine each login shell takes ~1s to initialize (rc files), so ~15 vars × ~1s ≈ 15s per test.

The test intent is pure string rendering — shelling out for env resolution inside a renderer is a latent footgun (tests flake on slow shells, CI variance).

## Fix shape

Pull the env resolution out of \`renderLaunchdPlist\` into \`bootstrapLaunchdJob\` (the only real caller that should do IPC). The renderer should only format what it's handed. Then the tests can pass \`env\` explicitly and stay under 1s.

Minimal alternative: give \`renderLaunchdPlist\` an optional \`resolveEnvVar\` dep (defaults to the real one), so tests can inject a no-op.

## Repro

\`\`\`
pnpm exec vitest run tests/breeze/breeze-daemon-launchd.test.ts
\`\`\`

serenakeyitan:
## Summary

\`renderLaunchdPlist\` was unconditionally calling \`readLoginShellEnvVarNames()\` and \`resolveLaunchdEnvVar()\` for every passthrough env var, and each of those \`spawnSync\`'s \`/bin/zsh -lc\`. On machines with slow login shells (rc files running nvm, starship, etc.), that's ~1s per subshell — three tests in \`breeze-daemon-launchd.test.ts\` were timing out at 5s after accumulating 15–25s of subshell init time.

## Fix

Add two optional injection points on \`LaunchdPlistInputs\`:

- \`resolveEnvVar?: (variable) => string | undefined\` — defaults to the production \`resolveLaunchdEnvVar\`, overridable in tests.
- \`loginShellVars?: readonly string[]\` — defaults to the result of \`readLoginShellEnvVarNames()\`, overridable in tests.

Production callers (\`bootstrapLaunchdJob\`) are untouched and still inherit real shell-resolved env. Tests pass a no-op resolver and an empty list, dropping the 3 failing tests from 5s+ timeout to 6ms total.

## Verification

\`\`\`
$ pnpm exec vitest run tests/breeze/breeze-daemon-launchd.test.ts
 Test Files  1 passed (1)
      Tests  10 passed (10)
   Duration  726ms
\`\`\`

Previously: 3 failed | 7 passed, 57s.

Closes #258

### DIFF
--- a/src/products/breeze/engine/daemon/launchd.ts
+++ b/src/products/breeze/engine/daemon/launchd.ts
@@ -31,6 +31,19 @@ export interface LaunchdPlistInputs {
   logPath: string;
   /** Extra env variables (HOME / PATH are added automatically). */
   env?: Record<string, string | undefined>;
+  /**
+   * How to fetch a passthrough env var that isn't already in `env`.
+   * Defaults to `resolveLaunchdEnvVar`, which shells out to
+   * `/bin/zsh -lc` — fine in production, but kills test perf on slow
+   * login shells. Pass a no-op or a stub in tests.
+   */
+  resolveEnvVar?: (variable: string) => string | undefined;
+  /**
+   * How to enumerate env vars from the user's login shell. Defaults to
+   * `readLoginShellEnvVarNames`, which shells out once. Tests pass an
+   * explicit list (or `[]`) to avoid the subshell.
+   */
+  loginShellVars?: readonly string[];
 }
 
 /** True on darwin with `launchctl` available. */
@@ -216,12 +229,16 @@ export function renderLaunchdPlist(inputs: LaunchdPlistInputs): string {
   for (const [key, value] of Object.entries(inputs.env ?? {})) {
     pushEnv(key, value);
   }
-  const passthroughVars = collectLaunchdPassthroughEnvVars({
-    ...process.env,
-    ...(inputs.env ?? {}),
-  });
+  const resolveVar = inputs.resolveEnvVar ?? resolveLaunchdEnvVar;
+  const passthroughVars = collectLaunchdPassthroughEnvVars(
+    { ...process.env, ...(inputs.env ?? {}) },
+    inputs.loginShellVars ?? readLoginShellEnvVarNames(),
+  );
   for (const variable of passthroughVars) {
-    pushEnv(variable, inputs.env?.[variable] ?? resolveLaunchdEnvVar(variable));
+    // Only fall back to the resolver for vars not already supplied;
+    // keeps tests and deterministic callers subprocess-free.
+    const supplied = inputs.env?.[variable];
+    pushEnv(variable, supplied ?? resolveVar(variable));
   }
 
   const environmentXml = envPairs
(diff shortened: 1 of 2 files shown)


bingran-you (approved):
LGTM — nice targeted fix.

Reviewed the diff and traced the callers:

- `bootstrapLaunchdJob` (`src/products/breeze/engine/daemon/launchd.ts:306`) passes neither injection point, so production preserves the old behavior via the two defaults.
- `renderLaunchdPlist` now only shells out when `loginShellVars` / `resolveEnvVar` aren't supplied, so tests (passing `[]` and a no-op) avoid `/bin/zsh -lc` entirely.
- Semantics of the supplied-vs-resolved precedence are preserved — `supplied ?? resolveVar(variable)` still falls back only on null/undefined, same as the pre-change `inputs.env?.[variable] ?? resolveLaunchdEnvVar(variable)`.
- `inputs.loginShellVars ?? readLoginShellEnvVarNames()` only invokes the costly default when the caller didn't provide one. Good.
- Checked gardener's `renderLaunchdPlist` — it doesn't do env resolution, so no sibling change needed.

The tree-sync check failure on this PR is unrelated (missing `TREE_REPO_TOKEN` secret in the `First-Tree Sync` workflow); the `ci` workflow itself is green.

One tiny thing, feel free to ignore: the two stubs in tests (`noResolve`, `noLoginShell`) are defined inside the outer `describe` but only used in three concrete tests — you could fold them into a shared `baseInputs` helper if the suite grows more, but at three call sites the explicit spread is perfectly readable. Not worth churning.

serenakeyitan:
Approving — clean, minimal perf fix.

**Looks good**
- Two optional DI hooks (`resolveEnvVar`, `loginShellVars`) on `LaunchdPlistInputs`, both defaulting to the existing production helpers. Production callers (`bootstrapLaunchdJob`) stay on real shell-resolved env.
- Signature matches `collectLaunchdPassthroughEnvVars(env, loginShellVars)` at `src/products/breeze/engine/daemon/launchd.ts:180`, so the new second arg threads through cleanly.
- Tests switch to `noResolve` + `[]`; I ran `pnpm exec vitest run tests/breeze/breeze-daemon-launchd.test.ts` locally — 10/10 passing in 13ms, matching the PR description.
- Doc comments on the new fields spell out why the defaults exist and when to override, which is exactly the context the next reader will need.

**Nit (non-blocking)**
- `inputs.loginShellVars ?? readLoginShellEnvVarNames()` evaluates the default eagerly every render, but `collectLaunchdPassthroughEnvVars` already has that same default, so you could drop the `??` here and let the helper handle it. Not worth another round-trip.

PR #264, opened 2026-04-22
fix(gardener): --dry-run should not mutate ~/.gardener/config.json

### LINKED ISSUE #263: bug(gardener/start): --dry-run writes ~/.gardener/config.json instead of just previewing
## Repro

\`\`\`bash
first-tree gardener start --tree-path /tmp/sandbox --code-repo acme/fake --dry-run
cat ~/.gardener/config.json
# => treePath: /tmp/sandbox, codeRepos: [acme/fake]
\`\`\`

## Expected

`--dry-run` should preview the config and boot plan without mutating any on-disk state.

## Actual

`src/products/gardener/engine/commands/start.ts:158` calls `writeDaemonConfig(config, env)` **before** the `if (flags.dryRun)` guard at line 168. So dry-run overwrites a live `~/.gardener/config.json` even though the daemon doesn't boot.

Discovered during an onboarding dry-run — I clobbered my own live config with a sandbox path. Recoverable (no daemon running), but a next restart would have booted pointing at a nonexistent tree path.

## Fix sketch

Move the `writeDaemonConfig` call below the `dryRun` branch, or accept a `cwd` override injection so tests/dry-runs don't touch `~/.gardener`.

bingran-you:
## Summary
- `writeDaemonConfig` ran before the `--dry-run` guard in `gardener start`, so a preview clobbered any existing `~/.gardener/config.json`.
- Move the write into the non-dry-run path; print a "preview, not written" header and the resolved config path when `--dry-run` is set.
- Update the dry-run test to assert the config file is left untouched (including an existing one).

Closes #263

## Test plan
- [x] `pnpm test tests/gardener/gardener-daemon.test.ts`
- [x] `pnpm typecheck`


serenakeyitan (approved):
Looks good to me. The fix keeps `--dry-run` from mutating `~/.gardener/config.json`, still shows the resolved config path in preview output, and adds coverage for leaving an existing config untouched. I also verified `pnpm test tests/gardener/gardener-daemon.test.ts` and `pnpm typecheck` locally.

bingran-you:
Self-review note: I checked the changed `gardener start --dry-run` path and the fix now previews the resolved config path without touching `~/.gardener/config.json`, including when an existing config is already present. I also verified `pnpm exec vitest run tests/gardener/gardener-daemon.test.ts` locally and found no blocking issues in this patch. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

serenakeyitan:
Thanks for the self-review note. I rechecked the latest PR head and reran `pnpm test tests/gardener/gardener-daemon.test.ts` plus `pnpm typecheck` locally; no further concerns from me.

bingran-you:
I rechecked the current PR head after the approval. The code change itself still looks fine, but the PR is blocked by `First-Tree Sync`: on commit `418ccd0`, the `Clone tree repo` step fails before gardener runs because the workflow cannot authenticate while cloning the private `agent-team-foundation/first-tree-context` repo and exits 128.

I didn’t push any source changes here because the PR diff is still only the `gardener start --dry-run` fix; the next step is to repair the workflow auth path for this repo (for example by refreshing the repo’s tree-repo credential/secret) and rerun the check.

serenakeyitan:
Confirmed — the PR merged at commit 418ccd0, so the gardener start --dry-run fix is in. The First-Tree Sync auth failure you flagged is orthogonal to this diff and needs a human to refresh the tree-repo credential/secret for this repo; flagging as out-of-scope for this PR. Closing out my side. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

Issue #265, opened 2026-04-22
tree-sync fails on fork PRs (TREE_REPO_TOKEN not injected)


serenakeyitan:
## Problem

The `tree-sync` workflow fails on every cross-repo PR because GitHub does not inject repo secrets into workflows triggered by fork PRs. Observed on PR #255 (run 24801666142):

\`\`\`
env:
  TREE_REPO_TOKEN: 
##[error]Input required and not supplied: token
\`\`\`

`TREE_REPO_TOKEN` is set on the repo (verified via `gh secret list`), but the value is blank at runtime because the PR head lives on `serenakeyitan/first-tree` (a fork).

This is intentional GitHub behavior — fork PRs don't get secret access for security reasons.

## Impact

Every fork PR shows a red `tree-sync` check that has nothing to do with the PR's actual content. Reviewers waste time investigating a known-broken gate.

## Fix

Skip `tree-sync` on fork PRs by adding a guard to the job:

\`\`\`yaml
jobs:
  tree-sync:
    if: github.event.pull_request.head.repo.full_name == github.repository
\`\`\`

The workflow will still run on in-repo branches (maintainer pushes, dependabot, etc.) where secrets are available. Fork PRs will show a skipped check rather than a failed one.

This guard belongs in both:
- `.github/workflows/first-tree-sync.yml` (this repo)
- `src/products/gardener/engine/install-workflow.ts` (`buildWorkflowYaml`), so every downstream repo that installs the workflow gets the same behavior.

## Alternatives considered

- `pull_request_target`: runs in base-repo context so secrets work, but checks out PR code — a real exfiltration risk for a workflow whose whole purpose is to process PR content.
- Accept the red check: noisy, trains reviewers to ignore CI.

## Out of scope

Enabling tree-sync verdicts *on* fork PRs would require a separate mechanism (e.g. a follow-up workflow_run job with access to secrets). Not addressing here.

serenakeyitan:
Fixed by #266 (merged 2026-04-22T21:37:49Z). tree-sync now skips cleanly on fork PRs via `github.event.pull_request.head.repo.full_name == github.repository` guard. The in-repo 'Bad credentials' problem is tracked separately in #268.

PR #266, opened 2026-04-22
fix(gardener/install-workflow): skip tree-sync on fork PRs

serenakeyitan:
## Summary

- Fork PRs can't access repo secrets, so `tree-sync` has been failing with `Input required and not supplied: token` on every cross-repo PR (e.g. #255).
- Adds a head-repo guard to the job `if:` so fork PRs show a skipped check instead of a red one.
- Mirrored in `buildWorkflowYaml` so every downstream repo that runs `first-tree gardener install-workflow` inherits the same behavior.

## Test plan

- [x] `pnpm exec vitest run tests/gardener/gardener-install-workflow.test.ts` — 21 pass (added assertion for the guard)
- [ ] Verify this PR itself gets a skipped (not failed) `tree-sync` check once CI runs

Refs #265

cc @bingran-you for review.


bingran-you (approved):
LGTM. Nice small, focused fix.

A few things I verified:

- The guard `github.event.pull_request.head.repo.full_name == github.repository` is the standard GitHub Actions pattern for detecting same-repo PRs, and it's correctly combined with the existing `first-tree:sync` label skip via `&&`.
- The change is mirrored in both the live workflow (`.github/workflows/first-tree-sync.yml`) and the `buildWorkflowYaml` template in `src/products/gardener/engine/install-workflow.ts:168`, so downstream repos running `first-tree gardener install-workflow` inherit it. Good catch keeping the two in sync.
- The comments correctly reflect each context's secrets (live workflow references `CLAUDE_CODE_OAUTH_TOKEN`, template references `ANTHROPIC_API_KEY`).
- The test assertion is minimal but sufficient for the template path.
- Best of all: this PR is itself a fork PR, and CI shows `tree-sync` as `SKIPPED` (not failed) — the fix is verified on the very change that introduces it.

Scope is appropriately narrow; the PR body correctly calls out that enabling tree-sync verdicts on fork PRs needs a separate `workflow_run`-based follow-up, which matches the approach I'd take too.

---

serenakeyitan:
Verified on this PR: `tree-sync` is now `skipping` instead of failing for the forked branch, and `pnpm exec vitest run tests/gardener/gardener-install-workflow.test.ts` passes locally (21/21). I do not see any blockers in this workflow guard change.

PR #270, opened 2026-04-22
fix: bypass actions/checkout REST auth for tree-sync

### LINKED ISSUE #268: infra: tree-sync fails with 'Bad credentials' despite valid TREE_REPO_TOKEN
## Problem

`tree-sync` workflow fails on every in-repo PR at the `Checkout tree repo` step with:

\`\`\`
env:
  TREE_REPO_TOKEN: ***     <-- secret is present (non-empty)
Retrieving the default branch name
Bad credentials *** https://docs.github.com/rest
##[error]Bad credentials *** https://docs.github.com/rest
\`\`\`

Observed on PR #264 (run 24803006416), but this affects every in-repo PR with a tree-sync check.

## What we tried

**Two distinct PATs, both validated OK from curl:**

| PAT type | `/user` | `/repos/.../first-tree-context` | Actions result |
|---|---|---|---|
| Fine-grained (scoped to `first-tree` + `first-tree-context`, `Contents: Read`, `Issues: R+W`) | 200 | 200 | Bad credentials |
| Classic (scope: `repo`) | 200 | 200 | Bad credentials |

Both were set on `agent-team-foundation/first-tree` via `gh secret set TREE_REPO_TOKEN --body -` (no trailing newline — `printf '%s'`).

The secret reaches the runner (log prints `TREE_REPO_TOKEN: ***`), so the value is non-empty. But `actions/checkout@v4`'s first REST call (to determine the default branch of `first-tree-context`) returns 401.

## Hypotheses not yet tested

1. **Secret value got corrupted at set time.** Try setting via the GitHub web UI (Settings → Secrets and variables → Actions) instead of the CLI.
2. **Org-level Actions policy.** `agent-team-foundation` may restrict which PATs can be used inside Actions workflows. Check org settings → Actions → General → PAT access for workflows.
3. **Fine-grained PAT org-approval pending.** When a fine-grained PAT is created against an org that hasn't pre-approved third-party tokens, repo API calls from `/user`+curl may succeed (owner self-access) while Actions runner requests (which may look like a different auth path) are rejected. Unlikely to explain classic PAT failing too, though.
4. **Switch to a GitHub App** for tree-repo access instead of a PAT. Longer-term answer regardless.

## Workaround in place

PR #266 adds a `if: github.event.pull_request.head.repo.full_name == github.repository` guard so fork PRs skip tree-sync cleanly. That doesn't help in-repo PRs like #264 though — those still hit this bug.

## Impact

In-repo PRs (bingran's branches, any maintainer push) are currently blocked from merging on repos where `tree-sync` is a required check, because the check never goes green. Admin-merge has been the de-facto workaround.

## Suggested next step

Try the GitHub web UI secret-set path first (5 min test). If that works, we have our answer and can document. If it still fails, investigate org-level Actions PAT policy before pulling in the GitHub App work.

/cc @bingran-you

bingran-you:
## Summary
- replace the generated cross-repo `actions/checkout` step with a direct `git clone` that uses `TREE_REPO_TOKEN`
- keep the source repo checkout on `actions/checkout`, but avoid its default-branch REST lookup for the tree repo
- update the install-workflow tests to pin the new YAML shape

## Testing
- `pnpm exec vitest run tests/gardener/gardener-install-workflow.test.ts`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test -- tests/gardener/gardener-install-workflow.test.ts` still trips an unrelated pre-existing failure in `tests/breeze/breeze-start.test.ts`

Fixes #268

### DIFF
--- a/.github/workflows/first-tree-sync.yml
+++ b/.github/workflows/first-tree-sync.yml
@@ -31,10 +31,28 @@ jobs:
           set -euo pipefail
           tree_repo_url="https://github.com/agent-team-foundation/first-tree-context.git"
           tree_repo_dir=".first-tree-cache/tree"
-          auth_header="$(printf 'x-access-token:%s' "$TREE_REPO_TOKEN" | base64 | tr -d '\n')"
           mkdir -p "$(dirname "$tree_repo_dir")"
-          git -c http.extraHeader="AUTHORIZATION: basic $auth_header" \
-            clone --depth 1 "$tree_repo_url" "$tree_repo_dir"
+          if git clone --depth 1 "$tree_repo_url" "$tree_repo_dir"; then
+            exit 0
+          fi
+
+          if [ -z "${TREE_REPO_TOKEN:-}" ]; then
+            echo "Anonymous tree clone failed and TREE_REPO_TOKEN is unset." >&2
+            exit 1
+          fi
+
+          rm -rf "$tree_repo_dir"
+          askpass_script="$RUNNER_TEMP/first-tree-git-askpass.sh"
+          cat >"$askpass_script" <<'EOF'
+          #!/bin/sh
+          case "$1" in
+            *Username*) printf '%s\n' 'x-access-token' ;;
+            *Password*) printf '%s\n' "$TREE_REPO_TOKEN" ;;
+          esac
+          EOF
+          chmod 700 "$askpass_script"
+          GIT_ASKPASS="$askpass_script" git clone --depth 1 "$tree_repo_url" "$tree_repo_dir"
+          rm -f "$askpass_script"
 
       - name: Setup Node
         uses: actions/setup-node@v4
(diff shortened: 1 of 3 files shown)


serenakeyitan (approved):
Looks good to me — the direct tree clone cleanly avoids the actions/checkout default-branch REST lookup while preserving the existing source checkout flow. I also ran pnpm exec vitest run tests/gardener/gardener-install-workflow.test.ts, pnpm typecheck, and pnpm build locally. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

bingran-you:
Local pass on current head `6cc0357`:
- `pnpm exec vitest run tests/gardener/gardener-install-workflow.test.ts`
- `pnpm typecheck`

I don't see a blocking issue in this three-file diff. The workflow swap from `actions/checkout` to an authenticated `git clone` is consistent with the generator/template updates and addresses the REST lookup path this PR is targeting.

GitHub still shows `REVIEW_REQUIRED`, and because this PR is authored by this account it still needs approval from another maintainer before merge.

PR #271, opened 2026-04-22
refactor(gardener/classifiers): extract verdict-parse shared module

serenakeyitan:
## Summary

- Extracts `parseVerdictJson`, `validateAndGroundNodes`, and related helpers out of `anthropic.ts` into a new `verdict-parse.ts` sibling.
- `anthropic.ts` re-exports both functions so existing test imports (`#products/gardener/engine/classifiers/anthropic.js`) keep working.
- No behaviour change — purely the refactor bingran asked for in #269 so the follow-up claude-cli classifier PR is purely additive.

## Test plan

- [x] `pnpm exec vitest run tests/gardener/gardener-anthropic-classifier.test.ts tests/gardener/gardener-comment.test.ts` → 125/125
- [x] `pnpm typecheck` clean
- [ ] CI on this PR

Refs #269

cc @bingran-you — prep PR per your review guidance. The claude-cli classifier will stack on top of this once it lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #269, opened 2026-04-22
gardener/comment: use local Claude Code auth instead of requiring ANTHROPIC_API_KEY


serenakeyitan:
## Problem

The Anthropic classifier shipped in #257 authenticates via `ANTHROPIC_API_KEY` against `api.anthropic.com/v1/messages`. On a developer machine that already has Claude Code installed and logged in, this is redundant friction:

- Claude Code's OAuth token is already in the keychain / `~/.claude/`
- Without `ANTHROPIC_API_KEY` exported, gardener silently falls back to the `INSUFFICIENT_CONTEXT/low` sentinel on every PR — the plumbing runs but the judgment is dead
- Users assume \"it's a local CLI and I'm logged in, why is it not working\" (confirmed in-session)

## Impact

Every `/gardener-comment` invocation on a machine with Claude Code but no exported API key produces low-signal verdicts. Observed live: dry-run against `paperclipai/paperclip` flagged 51 PRs as `INSUFFICIENT_CONTEXT/low`. If run for real it would have posted 51 useless comments.

## Proposed fix

Swap `createAnthropicClassifier`'s raw fetch for the Claude Agent SDK (`@anthropic-ai/claude-agent-sdk`), which picks up local Claude Code auth automatically. Fall back to `ANTHROPIC_API_KEY` when the SDK can't find a local session (CI, headless hosts).

Alternative considered: shell out to `claude -p \"<prompt>\"`. Works today, but adds a hard dependency on the `claude` binary being on `PATH` and is slower (cold-start per call).

## Acceptance

- [ ] `first-tree gardener comment` works on a fresh machine with Claude Code logged in, no `ANTHROPIC_API_KEY` set
- [ ] Existing `ANTHROPIC_API_KEY` path still works (CI, servers without Claude Code)
- [ ] Startup diagnostic tells the user which auth method was selected

## Related

- #257 — built-in Anthropic classifier (merged)
- #255 — refuse-to-post when no classifier injected (stacked on #257)

serenakeyitan:
## Design sketch — shell out to `claude -p`

### Why gardener currently needs `ANTHROPIC_API_KEY`

[...]

PR #272, opened 2026-04-22
feat(gardener/classifiers): add claude-cli classifier with auth-failure fallback

serenakeyitan:
## Summary
- Addresses #269: adds a claude-cli classifier so gardener can use the local `claude` OAuth session instead of requiring a separate `ANTHROPIC_API_KEY`.
- Adds `selectClassifier` with bingran's selection order: `GARDENER_CLASSIFIER` override → `claude` on PATH → `ANTHROPIC_API_KEY` → fail-closed.
- Transparent fallback: if claude-cli throws an auth-looking error AND `ANTHROPIC_API_KEY` is set, retry on the api-key path and warn once.

Stacked on #271 (refactor extracting the shared verdict-parse module). Once #271 merges, this PR is purely additive.

## Selection order (from #269 discussion)
1. `GARDENER_CLASSIFIER` env (`claude-cli` | `anthropic-api` | `none`) — CI/test escape hatch
2. `claude` binary on PATH → claude-cli (local dev path)
3. `ANTHROPIC_API_KEY` set → anthropic-api (CI path)
4. Neither → fail-closed (comment will not post, per #255)

## Test plan
- [x] `pnpm exec vitest run tests/gardener/gardener-classifier-select.test.ts tests/gardener/gardener-claude-cli-classifier.test.ts` → 16/16 pass
- [x] `pnpm typecheck` clean
- [ ] Manual smoke: run `first-tree gardener comment --pr N --repo owner/name` locally with `claude` authenticated and no ANTHROPIC_API_KEY set
- [ ] Manual smoke: run with `GARDENER_CLASSIFIER=none` and confirm no comment posted

Refs #269


serenakeyitan:
I think the new classifier selector still misses the macOS daemon path. `runGardener` now honors `GARDENER_CLASSIFIER`, but `first-tree gardener start` only persists `ANTHROPIC_API_KEY` and `GARDENER_CLASSIFIER_MODEL` into the launchd plist in `src/products/gardener/engine/commands/start.ts`. That means a launchd-backed daemon ignores a forced override like `GARDENER_CLASSIFIER=none`, `anthropic-api`, or `claude-cli`, so long-running sweeps behave differently from direct `gardener comment` invocations and the detached-spawn path.

Can we thread `GARDENER_CLASSIFIER` through the launchd env and add a regression test for the plist environment payload? This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

bingran-you (changes requested):
Requesting changes for two blocking issues:

1. `tests/gardener/gardener-classifier-select.test.ts:106` breaks `pnpm typecheck`. The new test imports `{ default: _ }` from `select.ts`, but that module has no default export. On this branch `pnpm typecheck` fails with `TS2339: Property 'default' does not exist ...`, so the release gate can't go green.

2. The new auto-selection path in `src/products/gardener/cli.ts:111-119` plus `src/products/gardener/engine/classifiers/select.ts` / `claude-cli.ts` makes `runGardener comment` depend on the host `claude` install and inherited `ANTHROPIC_API_KEY`. On a machine with `claude` on PATH, the existing `gardener-comment` dispatch tests now fail because the subprocess takes the claude-cli path and exits `Invalid API key · Fix external API key`, which `classifyFailure()` does not treat as auth-related, so the Anthropic fallback never fires. I reproduced that locally with `env ANTHROPIC_API_KEY=sk-test claude -p ...`, and `pnpm exec vitest run tests/gardener/gardener-comment.test.ts tests/gardener/gardener-classifier-select.test.ts tests/gardener/gardener-claude-cli-classifier.test.ts` fails in the two `runGardener ... built-in classifier` cases. Either the claude subprocess needs its env scrubbed so "local session auth" is actually local-session-first, or this path needs additional injection/coverage so it stays deterministic.

serenakeyitan:
Addressed both blockers in 39b51bc:

**1. `pnpm typecheck` break.** Removed the dead `{ default: _ }` import in `tests/gardener/gardener-classifier-select.test.ts:106`. Typecheck clean.

**2. `runGardener comment` taking claude-cli path in existing tests + "Invalid API key" not mapped to auth_failed.**

Root cause was deeper than just the missing stderr signal: claude-cli was inheriting `ANTHROPIC_API_KEY` from the parent env, so "local session first" wasn't actually local-session-first — with a test key like `sk-test`, claude used it and errored out.

Two-part fix:
- **Scrub `ANTHROPIC_API_KEY` from subprocess env** in `runClaude` — now the `claude` subprocess genuinely prefers local OAuth. Added test coverage that captures the `spawn` opts.env and asserts the key is absent.
- **Expand auth signals** to include `invalid api key`, `invalid_api_key`, `fix external api key`. Even if the host env still leaks the key somehow, the fallback path now fires correctly. Added a test with the exact stderr string you reproduced locally.
- The two existing `runGardener ... built-in classifier` tests in `gardener-comment.test.ts` now pin `GARDENER_CLASSIFIER=anthropic-api` (the designed-in escape hatch) so they're deterministic regardless of whether `claude` is on PATH.

Full suite: `pnpm exec vitest run` → 1110 passed / 51 skipped / 0 failed. `pnpm typecheck` clean.

Note: had to force-push because the branch was rebased past `43b50cb` (which landed via squash-merge of #271), so the old SHA couldn't be fast-forwarded. No content was overwritten.

serenakeyitan:
Addressed on `aab9e47`.

- Threaded `GARDENER_CLASSIFIER` through the launchd env in `first-tree gardener start`, with regression coverage for the rendered plist payload.
- Scrubbed inherited `ANTHROPIC_API_KEY` from the `claude` subprocess env so the claude-cli path stays local-session-first, and widened auth detection to treat `Invalid API key · Fix external API key` as retryable auth failure.
- Removed the stray default-import line in `tests/gardener/gardener-classifier-select.test.ts` that was breaking `pnpm typecheck`.
- Forced the two built-in classifier dispatch tests onto `GARDENER_CLASSIFIER=anthropic-api` so they stay deterministic even when `claude` is installed on the host.

Validation:
- `pnpm exec vitest run tests/gardener/gardener-classifier-select.test.ts tests/gardener/gardener-comment.test.ts tests/gardener/gardener-claude-cli-classifier.test.ts tests/gardener/gardener-daemon.test.ts`
- `pnpm typecheck`
- `pnpm build`
- `pnpm release:check` still reports pre-existing unrelated failures in `tests/breeze/breeze-daemon-runner.test.ts`, `tests/breeze/breeze-start.test.ts`, and two `tests/e2e/cli-e2e.test.ts` cases.

bingran-you (approved):
Rechecked at 39b51bc. The typecheck break is fixed, the claude-cli path now scrubs inherited ANTHROPIC_API_KEY and treats the invalid-key stderr shape as auth-related, and the launchd env now persists GARDENER_CLASSIFIER with regression coverage. I reran `pnpm typecheck`plus`pnpm exec vitest run tests/gardener/gardener-comment.test.ts tests/gardener/gardener-classifier-select.test.ts tests/gardener/gardener-claude-cli-classifier.test.ts tests/gardener/gardener-daemon.test.ts`; both passed locally. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

Issue #277, opened 2026-04-22
gardener + breeze: end-to-end onboarding flow (drift → per-proposal issues → assignee's breeze picks up)


serenakeyitan:
## The target UX

A single smooth flow so new users can onboard gardener + breeze together:

```
source repo activity (PR merges, new PRs)
        │
        ▼
gardener sync  ──►  detects drift between source repo and tree
        │
        ▼
gardener opens ONE tree-repo issue per proposal  ──►  assigns NODE owners
        │
        ▼
GitHub notification fires for each assignee
        │
        ▼
each assignee's breeze sees the notification on their laptop
        │
        ▼
breeze dispatches to the right agent to draft the tree update
        │
        ▼
agent opens a tree PR  ──►  gardener-respond handles review feedback
```

**Key shape:** routing by GitHub assignment. If gardener assigns @bingran-you on an issue, bingran's breeze picks it up. If it assigns @serenakeyitan, serena's breeze picks it up. No central queue, no shared server — assignment is the routing table.

## Why this matters

Today you have to know the CLI, know which subcommand to run, when to run it, and manually babysit the tree-update flow. With the above, onboarding is:

1. Install first-tree
2. Run `/gardener-onboarding` once
3. Start breeze
4. Done — drift gets detected, issues get filed against the right owner, breeze routes to the right agent, tree PRs get reviewed and merged

Each person runs their own breeze against their own notifications. No coordination needed.

## Current state vs. target state

| Step | Today | Target |
|---|---|---|
| Sync detects drift | ✅ `gardener sync` works | ✅ keep |
| Sync outputs | Opens ONE big PR with all edits (`--apply`) | One issue per proposal, assigned to NODE owners |
| Owner assignment on sync | ❌ `--assign-owners` doesn't apply to sync | ✅ per-proposal owner assignment |
| Breeze handles tree-repo issue notifications | ❓ unclear — handlers exist for source-repo PR/issue, not tree-repo issue assignment | ✅ assignment on tree issue → breeze dispatches \`<whatever writes the tree node>\` |
| Merge→issue branch on comment | ✅ works with \`--assign-owners\` + \`TREE_REPO_TOKEN\` | ✅ keep (and align with sync's per-proposal issue format) |
| Onboarding skill | \`/gardener-onboarding\` covers config + install-workflow + breeze install | Add a \"connect the pipes\" step so the above flow works end-to-end |

### 1. `gardener sync --open-issues` mode

[...]

PR #278, opened 2026-04-22
feat(gardener): add `sync --open-issues` mode (#277 gap 1)

serenakeyitan:
Refs #277 — first of 4 PRs closing the gardener + breeze onboarding pipeline described there.

## Summary

Adds `gardener sync --open-issues`: instead of bundling all drift proposals into one tree PR (`--apply`), open **one tree-repo issue per proposal**, assigned to the affected node's owners. The assignee's breeze (on their own laptop) picks up the GitHub notification and — in the gap-2 follow-up — dispatches an agent to draft the tree PR. GitHub assignment is the routing table.

**Issue shape:**
- Deterministic state marker: `<!-- gardener:sync-proposal · proposal_id=<sha256(path+body)[:12]> · source_sha=<short> · node=<path> -->`
- Labels `first-tree:sync-proposal` + `gardener`; `needs-owner` when falling back to the tree-repo default owner
- Body: rationale + proposed node content verbatim so the drafting agent has everything it needs

**Idempotency:** re-runs search for an open issue with the same `proposal_id` and skip if found.

**Auth:** requires `TREE_REPO_TOKEN`. Mutually exclusive with `--apply`. Honors `--dry-run`.

**Robustness:** retries without `--label` on 422 (repo hasn't seeded labels), and without `--assignee` on 422 (login isn't a collaborator), matching `comment`'s existing pattern.

## Design notes

Widens `ShellRun` in `src/products/tree/engine/runtime/shell.ts` to accept optional `env?`. Backward compatible — tree's own runner ignores it. Gardener's sync uses it to scope `gh issue create` to `GH_TOKEN=TREE_REPO_TOKEN` (same pattern as `handleMergedIssue`).

Skipped the DRY-ing of `createTreeIssue` between `handleMergedIssue` and this new path — that's a separate refactor PR. Keeps blast radius tight.

## Out of scope (tracked on #277)

- Align `comment`'s merge→issue body with this format (gap 3)
- Breeze dispatch rule for tree-issue assignment + `gardener draft-node` subcommand (gap 2)
- `/gardener-onboarding` walkthrough update (gap 4)

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm vitest run tests/gardener` — 298/298 pass (10 new in `sync-open-issues.test.ts`)
- [ ] CI green
- [ ] Manual E2E on a real tree repo with a seeded proposal (follow-up after merge)

### DIFF
--- a/src/products/gardener/engine/respond.ts
+++ b/src/products/gardener/engine/respond.ts
@@ -94,7 +94,7 @@ export interface ShellResult {
 export type ShellRun = (
   command: string,
   args: string[],
-  options?: { cwd?: string; input?: string; timeout?: number },
+  options?: { cwd?: string; input?: string; timeout?: number; env?: NodeJS.ProcessEnv },
 ) => Promise<ShellResult>;
 
 export interface RespondDeps {
--- a/src/products/gardener/engine/sync.ts
+++ b/src/products/gardener/engine/sync.ts
@@ -1437,8 +1437,6 @@ async function existingProposalIssueUrl(
       treeSlug,
       "--state",
       "open",
-      "--label",
-      "first-tree:sync-proposal",
       "--search",
       `proposal_id=${proposalId} in:body`,
       "--json",
--- a/src/products/gardener/engine/sync.ts
+++ b/src/products/gardener/engine/sync.ts
@@ -1,4 +1,5 @@
 import { execFile } from "node:child_process";
+import { createHash } from "node:crypto";
 import {
   existsSync,
   mkdirSync,
@@ -29,15 +30,21 @@ import type {
 
 const execFileAsync = promisify(execFile);
 
-export const SYNC_USAGE = `usage: first-tree gardener sync [--tree-path PATH] [--source ID] [--propose] [--apply] [--dry-run]
+export const SYNC_USAGE = `usage: first-tree gardener sync [--tree-path PATH] [--source ID] [--propose] [--apply | --open-issues] [--dry-run]
 
 Detect drift between a Context Tree and the source repo(s) it describes.
-Runs in three phases controlled by flags:
+Runs in three terminal modes controlled by flags:
 
-  default        Detect drift only. Prints a summary. Exits 0.
-  --propose      Detect + write proposal files under .first-tree/proposals/.
-  --apply        Detect + propose + write new tree files, commit, push,
-                 and open a PR labeled \`first-tree:sync\`.
+  default         Detect drift only. Prints a summary. Exits 0.
+  --propose       Detect + write proposal files under .first-tree/proposals/.
+  --apply         Detect + propose + write new tree files, commit, push,
+                  and open a PR labeled \`first-tree:sync\`.
+  --open-issues   Detect + propose + open ONE tree-repo issue per
+                  proposal, assigned to the affected node's owners. The
+                  issue carries a \`first-tree:sync-proposal\` label and a
+                  state marker so re-runs skip proposals that already
+                  have an open issue. Mutually exclusive with --apply.
+                  Requires TREE_REPO_TOKEN.
 
 First-run policy: if a binding has no \`lastReconciledSourceCommit\`, the
 command traces history back to the initial commit (capped at 500 commits
@@ -52,7 +59,10 @@ Options:
   --source ID         Only sync a single bound source by sourceId
   --propose           Write proposal files under .first-tree/proposals/
   --apply             Apply proposals, open PR (implies --propose)
-  --dry-run           With --apply, skip \`git push\` and \`gh pr create\`
+  --open-issues       Open one tree-repo issue per proposal instead of
+                      one bundled PR (implies --propose). Conflicts with --apply.
+  --dry-run           With --apply / --open-issues, skip network writes
+                      (\`git push\`, \`gh pr create\`, \`gh issue create\`)
   --help              Show this help message
 `;
 
@@ -66,7 +76,7 @@ export interface SyncDeps {
 async function defaultShellRun(
   command: string,
   args: string[],
-  options: { cwd?: string; input?: string; timeout?: number } = {},
+  options: { cwd?: string; input?: string; timeout?: number; env?: NodeJS.ProcessEnv } = {},
 ): Promise<ShellResult> {
   try {
     const { stdout, stderr } = await execFileAsync(command, args, {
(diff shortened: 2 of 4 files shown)


bingran-you (commented):
Nice scoping — pure helpers (`computeProposalId`, `buildSyncProposalBody`) extracted for testability, deterministic idempotency marker, graceful label/assignee fallbacks, and the `ShellRun` widening kept minimal. Couple of things I'd like to push back on before merging — mostly around what happens when the orchestration doesn't land on a happy path.

### Gaps worth addressing

**1. `needs-owner` is silently dropped when the author assignee is a non-collaborator.** In `runOpenIssuesMode` (`src/products/gardener/engine/sync.ts:~1505–1515`):

```ts
let assignees = owners;           // owners may be [authorLogin] only
let needsOwner = false;
if (assignees.length === 0) { … needsOwner = true; }
```

If the affected node has no `owners:` and `resolveProposalOwners` returns just the source-PR author, `assignees = [authorLogin]` and `needsOwner` stays `false`. When the author isn't a tree-repo collaborator, the 422-retry strips `--assignee` — and the issue ends up with *no* assignee *and* no `needs-owner` label. That's the exact case this fallback was built for, but the triage signal gets lost. Suggest: when the assignee-stripping retry fires, add `needs-owner` to the label set (or pre-classify "author-only" assignees as `needsOwner = true`).

**2. `resolveTreeDefaultOwner` returns `.owner.login`, which is usually an org.** For most tree repos that's an organization slug, which GitHub rejects as an issue assignee (issues can only be assigned to users). The default-owner fallback will therefore almost always trigger the 422-retry and open an unassigned issue — so the "assign to tree-repo default" branch silently no-ops in the common case. Options: pick from `.github/CODEOWNERS` fallback, read a maintainer-configurable `gardener.defaultAssignee`, or drop the attempt and just label `needs-owner` when there are no valid owners.

**3. No unit test for `runOpenIssuesMode` itself.** `shellRun` is already injectable, so a mock-based test for the happy path + the 422-retry branches + idempotency-hit would be straightforward and would cover the bits most likely to regress. The current 10 tests cover the pure helpers well, but the orchestration — which is the risky part — isn't exercised.

### Smaller notes (non-blocking)

- **Search-based idempotency has indexing lag.** `gh issue list --search "proposal_id=… in:body"` relies on GitHub's search index, which can lag minutes behind issue creation. Back-to-back `sync --open-issues` invocations can double-file. If that matters, a local `.first-tree/open-issues.json` cache of recent `proposal_id → url` would belt-and-braces the search path.
- **422 regex is loose.** `/label|422|unprocessable/i` will match unrelated 422s (oversized title, forbidden chars). The retries are additive-removals so it won't make things worse, but the log line will mislead about what was fixed.
- **Usage text** says "three terminal modes" but lists four; `--apply and --open-issues are mutually exclusive` exit code is 2 while other usage errors return 1 — intentional?
- **`runOpenIssuesMode` export** is only needed if tests or callers consume it; if not, keeping it module-private would shrink the surface.
- **Dry-run `created` counter** increments before any creation error could occur — that's correct for a preview, just noting it for anyone reading the numbers.

Happy to approve once #1 is addressed (or acknowledged as a follow-up on #277); #2 and #3 I'd take as separate PRs.

---

serenakeyitan:
I found two blocking edge cases in `src/products/gardener/engine/sync.ts` before this should merge:

- `existingProposalIssueUrl` still filters with `--label first-tree:sync-proposal` (`src/products/gardener/engine/sync.ts:1431`). If the repo hasn't seeded labels, issue creation falls back to an unlabeled issue, and the next run won't find it even though the `proposal_id` marker is present. That breaks the advertised idempotency and can create duplicates.
- The `gh issue create` retry sequence re-enables labels after a label failure (`src/products/gardener/engine/sync.ts:1576`). If labels are missing *and* the assignee isn't a collaborator, the second retry uses `withLabels: true` again, so the issue still can't be created even though a no-label / no-assignee fallback should succeed.

I’d fix those two cases before merging.

serenakeyitan:
Thanks for the careful read. Addressed #1 in 3010c81 — when the 422 assignee-strip retry fires, we now flip `needsOwner = true` and append `needs-owner` to the label set before re-issuing, with a secondary label-strip retry if the tree repo hasn't seeded labels. That's the case this fallback was built for, agreed the triage signal shouldn't get lost.

On #2 and #3, taking your suggestion to defer as follow-ups on #277 — I'll file:

- **#2** `resolveTreeDefaultOwner` returning an org: will drop the GET-repo-owner attempt and instead rely on CODEOWNERS / a configurable `gardener.defaultAssignee` / `needs-owner` fallback. Since the org-as-assignee branch already no-ops via the 422-retry (and now correctly lands `needs-owner`), this is no longer a correctness issue, just a silent-dead-code issue.
- **#3** mock-based orchestration tests for `runOpenIssuesMode` (happy path, 422-retries, idempotency hit).

On the smaller notes:
- Local `.first-tree/open-issues.json` cache for search-index lag — will include in the #3 follow-up.
- 422 regex tightening — will tighten alongside #3 tests.
- Usage text "three" → "four" modes + exit-code 1 vs 2 consistency — fixing in this PR (one-line nits).
- `runOpenIssuesMode` export is consumed by the test module, keeping it exported.
- Dry-run `created` counter — noted, leaving as-is.

Will push the usage-text nits shortly.

---

serenakeyitan:
Follow-ups filed: #280 (org-assignee cleanup), #281 (orchestration tests + cache + regex tightening). Also pushed usage-text nits in b27f14e.

bingran-you (approved):
Re-checked the follow-up commits: the assignee-strip retry now preserves the needs-owner triage signal and still falls back cleanly if labels are missing, which resolves the blocker I raised. I also reran the gardener test suite locally and it passed (298/298). This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

serenakeyitan:
Fixed the remaining idempotency blocker in `cfdc277`.

- `existingProposalIssueUrl` no longer filters on `first-tree:sync-proposal`, so reruns still find unlabeled fallback issues when the tree repo hasn't seeded labels.
- Added a focused `runOpenIssuesMode` test that proves we skip creation when an existing unlabeled proposal issue already matches the `proposal_id` marker.
- Validation: `pnpm exec vitest run tests/gardener` and `pnpm typecheck`.

bingran-you:
Re-checked cfdc277. Dropping the label filter from the existing-issue lookup closes the duplicate-issue path when label seeding failed, and the new runOpenIssuesMode regression test covers that unlabeled fallback directly. I reran pnpm exec vitest run tests/gardener (299/299) and pnpm typecheck; both passed. No further concerns from my side, and the earlier approval still stands. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

PR #282, opened 2026-04-22
feat(gardener): align comment merge-issue body with sync-proposal marker (#277 gap 3)

serenakeyitan:
## Summary

Third of four gaps under #277. Aligns the merged-PR issue-filing path in `gardener comment` with the dispatch-marker shape #278 (now on `main`) introduced for `sync --open-issues`, so breeze can match a single selector across both paths.

**What changed:**
- `buildTreeIssueBody` now prepends a `<!-- gardener:sync-proposal · proposal_id=… · source_sha=… · node=… · source=merged-pr · source_pr=<repo>#<n> -->` line as the first line of the tree-repo issue body.
- `computeMergedProposalId(sourceRepo, sourcePr)` — new exported helper — derives a stable 12-char SHA-256 slice from `"merged\n<repo>#<pr>"`. Deterministic for idempotency; a merged PR is one proposal-equivalent unit even when multiple tree nodes are cited.
- `sourceSha` added as an optional field on `BuildTreeIssueBodyInput` and `HandleMergedIssueInput`. Callers that don't pass it get `source_sha=unknown`; threading it from callers is a follow-up (non-blocking for dispatch).
- The marker picks the first cited tree node for `node=`; other nodes are still surfaced via the body's cc line.

## Why this shape

Sync's marker is single-node per proposal (one issue per node). The merged-PR path files one issue citing N nodes. Rather than forcing a multi-node marker shape (which would diverge from sync's), we keep `node=` single-valued (first-cited fallback) and add `source=merged-pr` + `source_pr=<repo>#<n>` so downstream can tell the paths apart and recover the full node list from the body.

Breeze's dispatch rule becomes a single regex — `/<!--\s*gardener:sync-proposal/` — firing on both paths.

## Test plan

- [x] `npx tsc --noEmit` passes.
- [x] `npx vitest run tests/gardener` — 303/303 pass (4 new marker tests).
- [x] New tests cover: marker as first line, `source_sha`/`node` unknown-fallbacks, first-node selection with multiple cited nodes, `proposal_id` determinism.
- [ ] E2E smoke on a real merged source PR (will run post-merge if bingran requests).

Refs #277.


bingran-you (approved):
LGTM. Clean execution of gap 3 — the marker shape is faithful to the sync variant and the design tradeoffs are spelled out where it matters.

**What I checked**

- Marker comparison: sync.ts:1369 emits `proposal_id · source_sha · node`; comment.ts:697-700 adds `source=merged-pr · source_pr=<repo>#<n>`. Both keep `node=` single-valued, so the `/<!--\s*gardener:sync-proposal/` selector fires uniformly while downstream can disambiguate via `source=`. Good shape.
- `computeMergedProposalId`: the `merged\n` prefix is sufficient to keep this disjoint from `computeProposalId` in sync.ts:1343 (which hashes `path\nbody`). Theoretical collision would require a node path literally equal to `merged` and a body that parses as `<repo>#<n>` — not worth defending against.
- Tests cover the load-bearing properties: marker is the first line, `source_sha`/`node` unknown-fallbacks, first-cited-node selection, and determinism across runs (and divergence across `sourcePr`).
- The PR description's acknowledgment that `sourceSha` threading from callers is a deferred follow-up matches what's in the diff (added to the input shapes; not yet wired in upstream callers). Marker degrades to `source_sha=unknown` cleanly in that interim.

**Minor / non-blocking**

- The `sourceSha` field is plumbed through `BuildTreeIssueBodyInput` and `HandleMergedIssueInput` but no caller passes it yet, so `source_sha=unknown` is the only value that ever lands in real issues until the follow-up. Worth a `TODO(#277)` near the call site in `handleMergedIssue` so the loose end is searchable.
- The `tree-sync` check is failing on auth (could not clone `first-tree-context`) — unrelated to this PR; safe to ignore for merge.

serenakeyitan:
Good catch — I addressed the loose end directly in `72b0db3` instead of leaving a TODO.

The merged-PR issue path now threads a real `sourceSha` into the dispatch marker, preferring the merged PR's `mergeCommit.oid` and falling back to `headRefOid` when needed, so this path no longer defaults to `source_sha=unknown` in normal operation. I also added a regression assertion in `tests/gardener/gardener-comment.test.ts` to lock that in.

Validated with `pnpm exec vitest run tests/gardener/gardener-comment.test.ts` and `pnpm typecheck`.

bingran-you:
Checked `72b0db3` locally. Threading `mergeCommit.oid` into `sourceSha` with the existing `headRefOid` fallback closes the only loose end I had called out, and the new assertion covers the merged-PR marker value directly. I reran `pnpm exec vitest run tests/gardener/gardener-comment.test.ts` and `pnpm typecheck`; both passed here. No further concerns from my side.

PR #286, opened 2026-04-23
fix(gardener): drop silent org-as-default-owner attempt in sync --open-issues

### LINKED ISSUE #280: gardener sync: drop org-as-assignee attempt in resolveTreeDefaultOwner
Follow-up on #278 (bingran review #2).

`resolveTreeDefaultOwner` returns the tree repo's `.owner.login`, which for most tree repos is an organization slug. GitHub rejects orgs as issue assignees, so that branch almost always triggers the 422-assignee-strip retry — after #278's fix it now correctly lands `needs-owner`, but the default-owner attempt is silent dead code in the common case.

**Options:**
1. Drop the `resolveTreeDefaultOwner` call entirely and rely on the `needs-owner` fallback when owners is empty.
2. Fall back to `.github/CODEOWNERS` parsing for the affected path.
3. Read a maintainer-configurable `gardener.defaultAssignee` from tree-repo config.

Refs #277.

serenakeyitan:
## Summary

Follow-up to #278 (bingran review #2).

`resolveTreeDefaultOwner` returned the tree repo's `.owner.login`, which is an organization slug for most tree repos. GitHub rejects orgs as issue assignees, so the call always failed and fell through to the 422-assignee-strip retry that already lands the issue unassigned with `needs-owner`. Net effect: silent dead code that made logs + issue body copy confusing.

This picks **Option 1** from #280 (the issue listed three): drop the call, treat empty-owners as `needs-owner` directly, update the issue-body copy to say "filing unassigned" instead of "assigned the tree-repo default".

The 422-assignee-strip retry from #278 still handles the case that actually matters (individual-user tree repos whose assignee isn't a collaborator).

Options 2 (CODEOWNERS parsing) and 3 (gardener.defaultAssignee config) can be filed as separate features if a real use case emerges — right now there's no evidence either was ever exercised.

## Test plan

- [x] Full gardener suite: 328/328 green
- [x] Typecheck clean
- [x] Existing `tests/gardener/sync-open-issues.test.ts` body-copy assertion still passes (it checks for "no \`owners:\`" and "needs-owner", which are preserved)

Closes #280
Refs #277

### DIFF
--- a/src/products/gardener/engine/sync.ts
+++ b/src/products/gardener/engine/sync.ts
@@ -283,6 +283,22 @@ function findExistingMemberDirByAuthorLogin(
   return matches[0] ?? null;
 }
 
+function dedupeOwners(owners: Iterable<string>): string[] {
+  const seen = new Set<string>();
+  const resolved: string[] = [];
+  for (const owner of owners) {
+    const cleaned = owner.replace(/^@+/, "").trim();
+    if (cleaned === "") continue;
+    const normalized = normalizeGitHubLogin(cleaned);
+    if (seen.has(normalized)) continue;
+    seen.add(normalized);
+    resolved.push(cleaned);
+  }
+  return resolved;
+}
+
+// Apply path (writing a new NODE.md): use node owners and fall back to the
+// PR author so the newly drafted node isn't ownerless.
 function resolveProposalOwners(
   treeRoot: string,
   targetDir: string,
@@ -292,18 +308,18 @@ function resolveProposalOwners(
   if (authorLogin) {
     combined.push(authorLogin);
   }
+  return dedupeOwners(combined);
+}
 
-  const seen = new Set<string>();
-  const resolved: string[] = [];
-  for (const owner of combined) {
-    const cleaned = owner.replace(/^@+/, "").trim();
-    if (cleaned === "") continue;
-    const normalized = normalizeGitHubLogin(cleaned);
-    if (seen.has(normalized)) continue;
-    seen.add(normalized);
-    resolved.push(cleaned);
-  }
-  return resolved;
+// --open-issues path: assignees must come from NODE.md owners only. If the
+// node has no `owners:`, we want `needs-owner` to fire — we must not silently
+// fall back to the PR author (that would mask the signal that #280 aims to
+// surface).
+function resolveIssueAssignees(
+  treeRoot: string,
+  targetDir: string,
+): string[] {
+  return dedupeOwners(resolveNodeOwners(targetDir, treeRoot, new Map()));
 }
 
 interface CommitSummary {
@@ -1481,12 +1497,7 @@ export async function runOpenIssuesMode(
         continue;
       }
       const absDir = join(treeRoot, proposal.path);
-      const owners = resolveProposalOwners(
-        treeRoot,
-        absDir,
-        classified.pr.authorLogin,
-      );
-      let assignees = owners;
+      let assignees = resolveIssueAssignees(treeRoot, absDir);
       let needsOwner = assignees.length === 0;
 
       const body = buildSyncProposalBody({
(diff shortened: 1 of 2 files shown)


serenakeyitan:
Blocking note from review: `runOpenIssuesMode()` still derives assignees via `resolveProposalOwners()`, and that helper appends `classified.pr.authorLogin` even when the affected tree path has no `owners:`. In that case this PR still won’t "treat empty-owners as `needs-owner` directly": if the source PR author is a tree-repo collaborator, the issue will be auto-assigned to them, `needsOwner` stays false, and the body claims it was assigned to node owners.

That seems to violate both #277’s routing contract (assignment should come from `NODE.md` owners) and #280’s stated fix (`needs-owner` when owners is empty). I think `--open-issues` needs to compute assignees from node owners only, or split `resolveProposalOwners()` so the apply path can keep the PR-author fallback separately.

bingran-you (approved):
I pulled the diff locally and do not see blockers with this cleanup. `pnpm exec vitest run tests/gardener/sync-open-issues.test.ts` and `pnpm typecheck` both pass for me. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

serenakeyitan:
Follow-up pushed as 73bc7fa — confirms the blocking note above was a real gap.

**What changed**
- Split `resolveProposalOwners()` into two call-site-specific helpers:
  - `resolveProposalOwners` (apply path, writing a new `NODE.md`) keeps the PR-author fallback so a freshly drafted node isn't ownerless.
  - `resolveIssueAssignees` (`--open-issues` path) resolves `NODE.md` owners only — no PR-author fallback — so empty `owners:` correctly flows to `needs-owner` instead of silently auto-assigning the source-PR author.
- Added a regression test in `tests/gardener/sync-open-issues.test.ts` that seeds a `NODE.md` with `owners: []` and a non-null `authorLogin`, then asserts the filed issue carries no `--assignee`, carries the `needs-owner` label, and contains the "no `owners:`" / `needs-owner` body copy.

**Validation**
- 329/329 gardener tests green (was 328/328 — +1 regression test)
- Typecheck clean

The 422-assignee-strip retry from #278 still covers the separate case of individual-user tree repos whose node-owner assignee isn't a collaborator.

bingran-you (changes requested):
I rechecked `73bc7fa` and the empty-owners path looks fixed, but there is still one gap in the existing assignee-strip retry.

`runOpenIssuesMode()` builds `body` before the `gh issue create` retry path. If the initial create fails because a node owner is not a collaborator, the code flips `needsOwner = true`, drops `--assignee`, and retries unassigned, but it reuses the original `--body`. That means the retried issue can land with the `needs-owner` label while still saying it was assigned to node owners from the affected `NODE.md`.

I think this needs one more follow-up so the retry path stays internally consistent too: either rebuild the body after the assignee fallback, or defer body construction until the final assignee/needs-owner state is known.

serenakeyitan:
Follow-up pushed as 5c429ac — confirms the retry-body gap you called out.

**What changed**
- Refactored `runOpenIssuesMode()` so `buildArgs()` regenerates the `--body` on every call via a `currentBody()` closure that reads the live `assignees` / `needsOwner` state. When the assignee-strip retry fires, the replayed `gh issue create` now carries a body that says "no `owners:`" / "filing unassigned" — matching the `needs-owner` label it lands with — instead of the stale "assigned to node owners" copy.
- Regression test in `tests/gardener/sync-open-issues.test.ts`: NODE.md seeded with `owners: [not-a-collaborator]`, first `gh issue create` returns a "not a collaborator" 422, retry without `--assignee` succeeds. Asserts the retry's `--body` contains "no `owners:`" / "filing unassigned" and not "assigned to node owners".

**Validation**
- 330/330 gardener tests green (was 329/329 — +1 regression test)
- Typecheck clean

bingran-you (approved):
Re-reviewed `5c429ac`. The retry-body gap I flagged is resolved: `buildArgs()` now regenerates `--body` via the `currentBody()` closure on every call, so after the assignee-strip fallback flips `needsOwner=true` and clears `assignees`, the retried `gh issue create` carries body copy that matches the `needs-owner` label it lands with.

I pulled the branch locally — `pnpm exec vitest run tests/gardener/sync-open-issues.test.ts` is 13/13 green (including the new retry-body regression) and `pnpm typecheck` is clean. No further blockers from my side.

PR #290, opened 2026-04-23
fix(breeze): use ?participating=true on notifications (drop ?all=true)

### LINKED ISSUE #251: breeze: `?all=true` bypasses GitHub spam filter — drive-by reviews on unrelated repos
## Summary

breeze polls GitHub notifications with `?all=true&participating=false`, which bypasses GitHub's server-side spam/quality filter. Combined with the default/common `repos: [all]` config, this causes breeze to act on notifications that **GitHub explicitly did not surface to the user in the UI** — including mention-then-delete spam and other filtered events.

In a real incident on 2026-04-21, a user running breeze with `repos: [all]` posted AI-generated reviews on **~15 unrelated public OSS repos** within about one hour (e.g. `Chainlit/chainlit#2901`, `BuilderIO/agent-native#253`, `tyleracates/TikTakToe#6`). None of those notifications were visible in the user's GitHub inbox UI. Maintainers on at least two repos (`asvishnyakov` on Chainlit, `rmems` on Limen-Neural) publicly asked "who are you and why are you approving my PR".

## Root cause

Three calls in breeze hit the notifications endpoint with filter-bypassing params:

- `src/products/breeze/engine/daemon/poller.ts:177` — `/notifications?all=true`
- `src/products/breeze/engine/daemon/gh-client.ts:91` — `/notifications?all=true&participating=false&per_page=100`
- `src/products/breeze/engine/commands/poll.ts:557` — `/notifications?all=true`

`?all=true` returns **every** notification the server has ever generated for the user, including ones GitHub has already silently filtered from the UI (observable as `unread: false` + `last_read_at: null` + `subject_type: null` on the notification record).

## Forensic evidence from the incident

On 3 stranger-repo notifications pulled from `~/.breeze/inbox.json`:

```json
{"repo": "tyleracates/TikTakToe",  "reason": "mention", "unread": false, "last_read_at": null}
{"repo": "Chainlit/chainlit",       "reason": "mention", "unread": false, "last_read_at": null}
{"repo": "BuilderIO/agent-native",  "reason": "comment", "unread": false, "last_read_at": null}
```

`unread: false` + `last_read_at: null` is the signature of a server-filtered notification: never surfaced to the user, never read by the user, but retained in the server-side pool. GitHub community discussion [#162922](https://github.com/orgs/community/discussions/162922) documents the mention-then-delete spam pattern that produces these.

Timeline API for all three PRs shows **no** `@serenakeyitan` in any PR body, comment, or review — consistent with the trigger comment having been deleted.

## Impact

For any user running breeze with `repos: [all]`:

1. GitHub cannot protect them. Their UI is clean — breeze still acts.
2. Any spammer who can land a mention that triggers a `reason=mention` notification (even a deleted one) can weaponize the user's account to post drive-by AI reviews on arbitrary public repos.
3. The user is unaware until a maintainer complains or they audit their own comment history.

## Proposed fix

Multiple defense layers:

### Layer 1 (must): drop `?all=true` by default

Change to plain `/notifications` (returns only unread, filtered by GitHub) or `/notifications?participating=true` (returns only items where the user is a direct participant — author, assignee, review_requested, mention). `participating=true` is the right default for an autonomous reviewer agent.

### Layer 2 (should): treat `repos: - all` as a soft warning

If the config resolves to all-repos AND the API is still `?all=true`, log a loud warning at startup and require an explicit `allow_unfiltered: true` knob to proceed.

### Layer 3 (nice): author/sender allowlist

Optional config knob to only respond to notifications triggered by users on a known list (e.g. `trusted_actors: [bingranyou, ...]`).

## Refs

- GitHub community: [Suddenly got several scam notifications — can I turn off mention in repos I don't belong to?](https://github.com/orgs/community/discussions/162922)
- GitHub community: [Phantom Notification from a repository I don't have access to](https://github.com/orgs/community/discussions/177901)
- GitHub Notifications API: https://docs.github.com/en/rest/activity/notifications

## Severity

High. Any breeze user on `repos: [all]` is exposed. The user's account posts real reviews on third-party public code in their name, with no way to detect it from the GitHub UI.

serenakeyitan:
## Summary
- Switch all three `/notifications` callers (poller.ts, gh-client.ts, poll.ts) from `?all=true` (or `?all=true&participating=false`) to `?participating=true`
- `?all=true` bypasses GitHub's server-side spam filter; `?participating=true` restricts to notifications where the user is a direct participant (author, assignee, mention, review_requested) and respects the filter
- This is Layer 1 from #251; Layers 2 (loud warning on `repos: [all]`) and 3 (trusted_actors allowlist) are left as separate follow-ups

Closes #251.

## Background

On 2026-04-21 a breeze user running with `repos: [all]` posted AI-generated reviews on ~15 unrelated public OSS repos within about an hour. The trigger notifications (`unread: false` + `last_read_at: null` + `subject_type: null`) were server-filtered — never visible to the user's GitHub inbox UI, but surfaced to breeze via `?all=true`. See issue #251 for full forensic evidence.

## Test plan
- [x] `pnpm exec vitest run tests/breeze/` — 424 passed
- [x] New regression assertion in `breeze-daemon-poller.test.ts` pins the URL to `/notifications?participating=true`
- [x] Existing `breeze-daemon-gh-client.test.ts` assertion updated to `participating=true&per_page=100`

### DIFF
--- a/src/products/breeze/engine/commands/poll.ts
+++ b/src/products/breeze/engine/commands/poll.ts
@@ -74,7 +74,7 @@ function formatUtcIso(date: Date): string {
   return `${date.toISOString().slice(0, 19)}Z`;
 }
 
-/** Raw GitHub notification as returned by `/notifications?all=true`. */
+/** Raw GitHub notification as returned by `/notifications?participating=true`. */
 interface RawNotification {
   id?: string;
   reason?: string;
@@ -554,7 +554,9 @@ export async function runPoll(
   try {
     const stdout = gh.runChecked("fetch notifications", [
       "api",
-      "/notifications?all=true",
+      // See #251: `all=true` bypassed GitHub's spam filter. `participating=true`
+      // restricts to direct-participation notifications.
+      "/notifications?participating=true",
       "--paginate",
       "-H",
       "X-GitHub-Api-Version: 2022-11-28",
--- a/src/products/breeze/engine/daemon/gh-client.ts
+++ b/src/products/breeze/engine/daemon/gh-client.ts
@@ -88,7 +88,10 @@ export class GhClient {
       "read recent notifications",
       [
         "api",
-        "/notifications?all=true&participating=false&per_page=100",
+        // See #251: `all=true&participating=false` bypassed GitHub's server-side
+        // spam filter and let breeze act on mention-then-delete notifications.
+        // `participating=true` restricts to direct-participation notifications.
+        "/notifications?participating=true&per_page=100",
         "--paginate",
         "-H",
         "X-GitHub-Api-Version: 2022-11-28",
--- a/src/products/breeze/engine/daemon/poller.ts
+++ b/src/products/breeze/engine/daemon/poller.ts
@@ -174,7 +174,12 @@ export async function pollOnce(deps: PollOnceDeps): Promise<PollOutcome> {
   try {
     const stdout = gh.runChecked("fetch notifications", [
       "api",
-      "/notifications?all=true",
+      // `participating=true` restricts to notifications where the user is a
+      // direct participant (author, assignee, mention, review_requested) and
+      // respects GitHub's server-side spam filter. `?all=true` was previously
+      // used here but bypassed the filter, causing breeze to act on
+      // mention-then-delete spam surfaced to no one in the UI (#251).
+      "/notifications?participating=true",
       "--paginate",
       "-H",
       "X-GitHub-Api-Version: 2022-11-28",
--- a/tests/breeze/breeze-daemon-gh-client.test.ts
+++ b/tests/breeze/breeze-daemon-gh-client.test.ts
@@ -119,7 +119,7 @@ describe("GhClient.recentNotifications", () => {
     expect(tasks[0].kind).toBe("mention");
     // Assert the argv sent to gh contains the paginate + notifications endpoint.
     expect(ctl.calls[0].args).toContain("--paginate");
-    expect(ctl.calls[0].args).toContain("/notifications?all=true&participating=false&per_page=100");
+    expect(ctl.calls[0].args).toContain("/notifications?participating=true&per_page=100");
   });
 
   it("drops lines that predate the lookback window", async () => {
--- a/tests/breeze/breeze-daemon-poller.test.ts
+++ b/tests/breeze/breeze-daemon-poller.test.ts
@@ -37,7 +37,7 @@ interface ResponseMatcher {
   stderr?: string;
 }
 
-function makeStubGh(matchers: ResponseMatcher[]): GhClient {
+function makeStubGh(matchers: ResponseMatcher[]): GhClient & { calls: string[][] } {
   const calls: string[][] = [];
   const spawn = vi
     .fn()
@@ -64,7 +64,8 @@ function makeStubGh(matchers: ResponseMatcher[]): GhClient {
         output: [],
       };
     });
-  return new GhClient({ spawn });
+  const client = new GhClient({ spawn });
+  return Object.assign(client, { calls });
 }
 
 function mkBreezeDir(): { dir: string; inbox: string; activity: string } {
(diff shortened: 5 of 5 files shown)


bingran-you (approved):
## Review: LGTM with one small follow-up question

Thanks for the fast, focused Layer-1 fix. I walked through the three call sites, the tests, and the incident evidence in #251, and I'm ready to approve.

### What I verified
- All three `/notifications` callers are migrated consistently:
  - `src/products/breeze/engine/commands/poll.ts:559`
  - `src/products/breeze/engine/daemon/gh-client.ts:94`
  - `src/products/breeze/engine/daemon/poller.ts:182`
- No stragglers elsewhere in the repo (`rg "all=true"` is clean; the only remaining references are in the new explanatory comments).
- Inline comments clearly explain the "why" and point back to #251 — exactly the kind of surprise-preventing comment worth keeping.
- Regression tests pin the URL at both layers (`gh-client` unit + `poller` integration). Exposing `gh.calls` from `makeStubGh` via `Object.assign` is a bit unusual but reads fine for a test helper and is the minimum change needed to assert the argv.
- Local `pnpm exec vitest run tests/breeze/breeze-daemon-poller.test.ts tests/breeze/breeze-daemon-gh-client.test.ts` → 26/26 passing.

### Correctness note on the semantic shift
The old query (`?all=true` in poll.ts/poller.ts, `?all=true&participating=false` in gh-client.ts) returned "everything the server ever generated for this user." The new query (`?participating=true`) returns only author / assignee / mention / review_requested. For an autonomous reviewer agent this is the right default, and it's consistent with the "direct participation only" stance documented in #251. Worth being explicit that users who previously relied on breeze surfacing subscribed-repo activity (non-mention `subscribed`-reason notifications from repos they watch) will lose those — I think that's the intended trade-off, but good to mention in the release notes when this ships.

### One small question (non-blocking)
The body says "Closes #251" but also "Layers 2 (loud warning on `repos: [all]`) and 3 (trusted_actors allowlist) are left as separate follow-ups." Do you want to:
- keep "Closes #251" and open two fresh issues for Layers 2 and 3 before merge, or
- switch to "Addresses #251" so the parent stays open until the other two layers are tracked somewhere?

Either is fine — I just want to make sure Layers 2 and 3 don't get lost when this lands.

Approving. Thanks for the thorough write-up and forensic evidence on the original issue — made this a very easy review.

serenakeyitan:
I don't think we should merge this as-is.

Switching every notifications call to `?participating=true` does fix the stranger-spam path, but it also narrows breeze more than the current task model does. In particular, `src/products/breeze/engine/runtime/task-kind.ts` still treats `manual` notifications as actionable, and those are exactly the kind of notifications I'd expect to disappear once we stop asking for the broader notifications feed. The updated tests all pass, but they only pin the new endpoint; they do not prove we preserve the actionable reason set that breeze still routes today.

Given the note already captured on #251, I think the safer path is either:
- keep the broader notifications query and filter the bad signature explicitly (`unread === false` plus missing `subject.type`, with `last_read_at` carried through), or
- add fixtures that demonstrate `participating=true` does not drop any notification reasons we still intentionally process before we merge this.

bingran-you (changes requested):
Serena is right about the regression surface here. I re-checked the notification paths against the current code and live API data on this account, and `?participating=true` narrows breeze more than Layer 1 intended: it drops recent actionable read notifications entirely, not just the hidden stranger-spam signature from #251. On my current feed, `?participating=true` returned nothing while `?all=true` still contained recent `comment` / `mention` / `author` activity that breeze would route today. I pushed a concrete follow-up fix to `origin/fix/251-drop-all-true-notifications` at `c08399b`, but this PR points at a hidden head ref (`refs/pull/290/head`), so GitHub will not let maintainers update it directly. That patch restores the broad feed, drops only the conservative hidden-thread signature (`unread=false` with missing `subject.type`), adds regression coverage that keeps read actionable threads such as `manual`, passed the touched-suite tests (61/61), passed full `tests/breeze/` with `BREEZE_DIR` / `BREEZE_HOME` unset so the runner env does not leak into `breeze-start.test.ts` (428/428), and passed `pnpm typecheck`. Please cherry-pick `c08399b` or reopen from the visible branch and I will re-review immediately. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

Issue #299, opened 2026-04-23
gardener-onboarding: Scenario F/C doesn't demonstrate the issue → draft-node → PR chain


serenakeyitan:
Split from tester feedback on #291.

## Problem

Testers reading \`skills/gardener/SKILL.md\` cannot tell that the issue-first flow (gardener opens issue → owner's breeze receives notification → dispatches to gardener draft-node → draft-node opens fix PR) is the actual end-to-end routing mechanism.

Concrete confusions reported:

1. **Pull mode appears to skip the issue step.** Scenario C shows \`gardener start\` with no mention of \`--open-issues\` vs \`--apply\`. Testers run \`gardener sync --apply\` and see a PR land directly, concluding the issue-first flow doesn't exist in pull mode.
2. **No explicit draft-node step.** The pipeline diagram at lines 30–56 mentions \"breeze dispatches to the right agent (gardener draft-node)\" but Scenario F's smoke test doesn't walk through the draft-node invocation or show what the owner should expect to see on their laptop.
3. **Tester literally concluded** \"there's no 'issue → auto-PR' bot in the documented flow.\" That's the core promise of the system and it reads as missing.

## Fix

Rewrite Scenario C + Scenario F so:

- Scenario C defaults to \`gardener sync --open-issues\` (multi-owner case), with \`--apply\` documented as the no-routing shortcut.
- Scenario F walks the full chain with concrete commands: gardener files the issue → owner's breeze sees the notification → breeze dispatches draft-node → draft-node opens the PR → gardener respond handles review feedback.
- Each step has an observable \"you should now see X\" verification (inbox entry, notification on statusline, new PR URL).

## Acceptance

- A friend running the onboarding prompt on a fresh machine can follow Scenario F to a real PR without asking \"where did the issue go\" or \"what opens the PR.\"
- The \`--open-issues\` vs \`--apply\` decision is framed as a routing choice, not a hidden flag.

Related: #292 (viewerPermission auto-detect), #294 (--allow-repo in pull mode).

serenakeyitan:
Addressed in #300: Scenario F now walks through the full issue → `gardener draft-node` → tree PR chain, with per-step Expect/Checkpoint callouts and the pipeline diagram reproduced inline for localizing the broken arrow. Closing as shipped.

Issue #295, opened 2026-04-23
breeze: pull mode silently depends on GitHub watch subscription — add doctor/check


serenakeyitan:
Split from #291 item 4.

## Problem
Pull mode requires the user to be watching the source repo on GitHub — otherwise GitHub delivers no notifications, poller returns empty, nothing happens. Skill doesn't mention this. \`gh api /repos/<owner>/<repo>/subscription\` requires the \`notifications\` scope that default \`gh auth login\` doesn't grant.

## Fix ideas
- New \`breeze doctor\` (or \`breeze check-watches\`) verifies each repo in \`repos:\` has an active subscription; prints the \`https://github.com/<repo>/subscription\` URL when missing.
- \`gardener-onboarding\` skill should call this out explicitly with the Watch URL for the source repo.

## Acceptance
- A command exists to verify watch state per repo.
- Onboarding skill documents the watch requirement and shows the Watch URL.

serenakeyitan:
Obsolete after #290 + #300. Breeze now fetches with `?participating=true`, so assignees receive notifications via the issue assignment itself — no watch subscription required. Skill (#300) documents this explicitly under Scenario C. Closing.

Issue #294, opened 2026-04-23
gardener-onboarding skill doesn't mention --allow-repo requirement in pull mode


serenakeyitan:
Split from #291 item 3.

## Problem
Skill's pull-mode section says just \`first-tree breeze install && first-tree breeze start\`. In practice both commands require \`--allow-repo <tree>,<source>\` to actually deliver notifications.

## Fix
Update \`skills/gardener/SKILL.md\` pull-mode scenario to show the real command with \`--allow-repo\` populated from the tree + source slugs resolved earlier in the flow.

## Acceptance
- Skill's pull-mode step shows \`first-tree breeze start --allow-repo <tree>,<source>\` (or equivalent multi-flag form).
- Comment explains the flag is required, not optional.

serenakeyitan:
Addressed in #300: Scenario C Step 2 now explicitly documents `--allow-repo <owner>/<repo>` as required for pull-mode breeze, with an Expect callout showing the daemon refuses to start without it. Closing as shipped.

Issue #292, opened 2026-04-23
gardener-onboarding: auto-detect viewerPermission and recommend mode


serenakeyitan:
Split from #291 item 1.

## Problem
`gardener-onboarding` skill presents push-mode vs pull-mode as a judgment call. Skill could auto-detect the source repo's `viewerPermission` via `gh repo view <slug> --json viewerPermission` and recommend pull mode when it's READ.

## Fix
Update `skills/gardener/SKILL.md` onboarding flow so it auto-runs the permission check and branches into Scenario B (push) or Scenario C (pull) deterministically instead of asking the user.

## Acceptance
- Skill includes the `gh repo view --json viewerPermission` step before the scenario fork.
- Scenarios B/C text starts with "Detected when viewerPermission is {ADMIN,WRITE,MAINTAIN} / READ".

serenakeyitan:
Addressed in #300: Scenario C now opens with Step 0 running `gh repo view --json viewerPermission` and a routing-mode decision table that picks the right mode based on that output. Closing as shipped.

PR #300, opened 2026-04-23
docs(gardener): rewrite onboarding skill for pull-mode clarity

serenakeyitan:
## Summary

- Add **Step 0** before the scenario fork: run `gh repo view --json viewerPermission` and pick Scenario B (push) vs C (pull) deterministically instead of asking the user.
- Rewrite **Scenario C** with a routing-mode decision table (`sync --open-issues` vs `sync --apply`), the required `breeze start --allow-repo <tree>,<source>`, and an explicit GitHub watch-subscription verification step.
- Expand **Scenario F** with per-step "Expect:" / "Checkpoint:" callouts so testers can tell which arrow in the pipeline broke without guessing.

## Motivation

Testers onboarding gardener + breeze against paperclip reported three confusions (see #291):

1. Pull mode appeared to skip the issue-first flow because Scenario C didn't mention `--open-issues` — they ran `--apply` and concluded the issue → draft-node → PR chain didn't exist.
2. `breeze start` didn't work with the defaults alone — `--allow-repo` is required but wasn't shown.
3. The GitHub watch-subscription requirement was undocumented, so pull-mode installs silently delivered nothing.

## Refs

Addresses: #292 #294 #295 (skill side) #299
Umbrella: #291

## Test plan

- [ ] Have a tester run the canonical onboarding prompt against a new pair of repos with READ-only source and confirm they land in Scenario C with `sync --open-issues` without asking about routing mode.
- [ ] Confirm Scenario F's new checkpoints match what the tester actually sees (inbox entry, PR URL, issue labels).

### DIFF
--- a/skills/gardener/SKILL.md
+++ b/skills/gardener/SKILL.md
@@ -27,6 +27,26 @@ infer what you can, ask only when the inference is genuinely ambiguous.
 Always finish by restarting the breeze daemon with the new
 `--allow-repo` list so notifications flow.
 
+### Step 0 — auto-detect the deployment mode
+
+Before picking a scenario, resolve the source repo's permission level so
+you don't guess push vs pull:
+
+```bash
+gh repo view <source-slug> --json viewerPermission -q .viewerPermission
+```
+
+- `ADMIN` / `MAINTAIN` / `WRITE` → go to **Scenario B** (push mode,
+  lower latency, no daemon).
+- `READ` / `TRIAGE` / `NONE` → go to **Scenario C** (pull mode, local
+  daemon). The user cannot land a workflow file in the source repo;
+  don't even suggest it.
+- Command fails (repo not found, not logged in) → stop and ask the
+  user; do not silently fall back.
+
+If the tree repo doesn't exist yet, run **Scenario A** first, then come
+back to this step.
+
 ### The end-to-end pipeline (what onboarding is actually wiring)
 
 Gardener + breeze together form a routing pipeline. Scenarios A–E below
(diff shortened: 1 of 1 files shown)


bingran-you (commented):
Thanks — the Step 0 auto-detect, the routing-mode table in Scenario C, and the per-step Expect/Checkpoint callouts in Scenario F are a real improvement. The pipeline diagram reproduced inside Scenario F especially helps testers localize the broken arrow instead of guessing. A few things worth addressing before merge; I don't think any are blockers but the first one is a correctness concern.

**1. Scenario C Step 4 — the watch-subscription framing is misleading given `?participating=true`.**

Since #290, breeze hits `/notifications?participating=true` in all three callers (`poll.ts`, `gh-client.ts`, `poller.ts`). That filter returns only threads where the user is a direct participant (assignee / mention / review-requested / author). Two consequences for this step:

- For the routing flow Scenario F proves out: the assignee gets a participating notification from the assignment itself, regardless of whether they're Watching the tree repo. So the watch requirement isn't what makes the chain work.
- For source-repo activity: Watching alone won't surface anything through `?participating=true` either — the user still needs to be a direct participant on the source-repo item. Watching a third-party source repo you aren't participating in is a no-op under the current filter.

Also, note that pull-mode gardener itself doesn't consume GitHub notifications at all — `gardener sync` polls source via `gh` directly. So the sentence *"Pull mode relies on GitHub notifications, which only fire if the user is watching the source repo. … Without this, the poller returns empty and nothing ever happens"* conflates gardener's polling sweep with breeze's notifications poller, and frames watching as necessary+sufficient when under `?participating=true` it's neither.

Suggested rewrite direction: drop step 4 as written, and instead (a) in Scenario F's preconditions note that the assignee will receive a participating notification via GitHub assignment (no watch needed), and (b) if keeping any subscription guidance for the source repo, scope it to the specific case where the user *is* author/reviewer/assignee on source PRs and wants breeze to pick those up.

**2. Scenario B opener is out of sync with Step 0.**

Step 0 now gates routing on `gh repo view --json viewerPermission -q .viewerPermission`, but Scenario B still opens with `` `gh api repos/<codebase> -q '.permissions.push'` returns true. `` Since Step 0 is the single detection point, Scenario B could just say "Step 0 returned ADMIN / MAINTAIN / WRITE" — matching how Scenario C now opens. Otherwise testers have two different probes for the same fact.

**3. Nit — command-invocation consistency in Scenario C.**

Step 2 uses bare `gardener start`; Step 3 uses `first-tree breeze start`. Elsewhere in the doc (the CLI tables, Scenario F steps 3/4/6) the normalized form is `first-tree gardener <command>`. Aligning Scenario C on `first-tree gardener start ...` would match.

**4. Small clarity win in Scenario C's decision table.**

The `--open-issues` vs `--apply` distinction is great. One thing a first-time tester won't have context for: *why* issue-first is "the routing path". It's worth a one-liner in the table cell or right under it noting that `--open-issues` is what makes GitHub assignment the distributed queue each owner's local breeze pulls from (i.e. it's the only mode that exercises the draft-node dispatch that Scenario F validates). `--apply` skips the routing step entirely — Scenario F isn't applicable there.

**5. Very minor — Step 0.**

`Command fails (repo not found, not logged in) → stop and ask the user; do not silently fall back.` is good. Consider also calling out that `viewerPermission: null` on a public repo (common when not logged in even though the command "succeeds") should be treated as the fail branch, not as READ. In practice `.viewerPermission` is `READ` for any authenticated non-collaborator on a public repo, but if auth is missing entirely the field comes back as null and falls through the existing cases.

Nothing above blocks the docs value of this change — the rewrites are a clear win over the previous one-line Scenario C. Happy to approve once (1) is addressed or explicitly justified; the rest are optional polish.

serenakeyitan:
I found two doc mismatches while reviewing the onboarding rewrite and pushed a follow-up fix to this PR branch: Scenario C now matches the actual pull-mode CLI surface (`gardener sync --open-issues` is documented as a manual issue-first step, while automatic batched PRs use `gardener start --sync-apply`), and the notification guidance now points at the tree-repo issue path used by Scenario F instead of the source repo watch path. `pnpm validate:skill` passes after the edit. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

serenakeyitan:
Addressed all 5 review points in f00eaaf:

**1 (blocker)** — Scenario C's watch-subscription step dropped. Added a note under the start-breeze step explaining why watching isn't required under \`?participating=true\` (and that pull-mode gardener doesn't consume notifications at all, it polls source via \`gh\`). Also cleaned up Scenario F precondition + troubleshooting hints that referenced watching.

**2** — Scenario B opener now reads \"Step 0 returned ADMIN / MAINTAIN / WRITE\" to match C.

**3** — Scenario C commands normalized to \`first-tree gardener …\` / \`first-tree breeze …\`.

**4** — Decision-table row for \`--open-issues\` now explains it's the only mode that exercises draft-node dispatch (distributed queue via GitHub assignment) and notes Scenario F doesn't apply to \`--apply\`.

**5** — Step 0 now treats \`viewerPermission: null\` (public repo, unauthenticated gh) as the fail branch, not READ.

Thanks for the careful review — (1) especially was an outright error, not just framing; glad it got caught before the skill went to testers.

serenakeyitan:
Follow-up: breeze autonomously pushed 3610a89 while I was drafting my review-response. Rebased my changes on top and reconciled: kept breeze's correct observation that `--open-issues` is operator-run today (not a `gardener start` mode) and that batched-PR routing is `gardener start --sync-apply`. Kept my drop of the watch-subscription step (bingran's #1), since breeze's reframing of that step still left the misleading implication that watching gates the flow.

bingran-you (approved):
LGTM — all five points addressed cleanly in 1b2525d. Specifically:

1. **Watch-subscription framing (the correctness concern)** — fully resolved. Dropping Step 4's watch requirement and replacing it with the callout explaining `?participating=true` + that pull-mode gardener polls via `gh` rather than consuming notifications matches how the code actually behaves post-#290. Good catch also removing the parallel references from Scenario F's preconditions and the "most common breaks" list — those would have been confusing leftovers. The replacement common-break (`gh auth status` login not matching `owners:`) is a real failure mode worth calling out.
2. Scenario B opener now mirrors Scenario C's "Step 0 returned …" phrasing — single detection point, no drift.
3. Commands normalized to `first-tree <product> <command>`.
4. The decision-table row now makes explicit that `--open-issues` is what exercises the draft-node dispatch (distributed queue via assignment) and that Scenario F doesn't apply to `--sync-apply`. First-time testers will now know why they're picking this mode, not just what it does.
5. Step 0 now treats `viewerPermission: null` as the fail branch — the exact edge case that would have sent anonymous public-repo viewers down the pull-mode branch silently.

Approving.

Issue #297, opened 2026-04-23
breeze: add-repo command for ergonomic allowlist extension


serenakeyitan:
Split from #291 item 6. Related: #293.

## Problem
No way to extend the breeze daemon's allowlist without stopping it and re-starting with the full csv. Onboarding a second tree today means retyping all existing allow-repo entries.

## Fix
\`first-tree breeze add-repo <owner>/<repo>\` — edits \`~/.breeze/config.yaml\` \`repos:\`, restarts the daemon with the merged list as the new allowlist.

## Acceptance
- \`breeze add-repo <slug>\` works idempotently (no duplicates).
- Daemon restarts (or reloads, if #293 ships \`reload\`) with the new repo in the allowlist.
- \`breeze --help\` shows the new command.

serenakeyitan:
Blocked on a design question: breeze's \`config.yaml\` doesn't currently have a \`repos:\` field. The allow-list is CLI-arg only (\`--allow-repo\`) and isn't persisted anywhere the daemon reads on restart.

So \`breeze add-repo <slug>\` needs to either:

1. **Introduce a new \`repos:\` field in \`~/.first-tree/breeze/config.yaml\`** as the persistent source of truth, teach \`breeze start\` to merge it with \`--allow-repo\`, and \`add-repo\` just edits the file + restarts. Cleanest.
2. **Read the running daemon's allow-list** (from the runtime status file that \`breeze status\` already reads via \`runtime.get(\"allowed_repos\")\`), append, stop, and restart with the merged csv. No config change needed, but the logic lives entirely in \`add-repo\` and re-derives state from runtime every time.
3. **Persist the last \`--allow-repo\` csv** to disk on \`start\` (e.g. \`~/.breeze/last-allow-repo.txt\`), have \`add-repo\` read + edit + pass on restart.

Option 1 is the most durable; option 2 works but breaks if the daemon isn't running. Option 3 is the smallest change.

Deferring this until we pick one. Related fix for #293 shipped in #301 (loud failure on divergent-allowlist start).

serenakeyitan:
## Design options (follow-up to earlier comment)

Three concrete shapes for `breeze add-repo`:

### Option A: New persistent `repos:` field in `config.yaml`
[...]

Issue #291, opened 2026-04-23
Onboarding friction: gardener + breeze setup for a READ-only source repo


serenakeyitan:
## Context

Onboarded gardener + breeze for a new tree (`serenakeyitan/paperclip-tree`) watching a source repo (`paperclipai/paperclip`) where the user has READ-only access. Collecting friction points encountered so the next onboarding is smoother.

### 1. Skill doesn't detect permission level and recommend mode
`gardener-onboarding` presents push vs pull as a judgment call. The skill could auto-detect `viewerPermission` on the source repo via `gh repo view --json viewerPermission` and recommend pull mode when permission is READ. Would have saved a manual check.

### 2. `breeze start` silently no-ops if daemon already running with a different `--allow-repo`
Daemon was already running (pid 97184) with `allowed repos: serenakeyitan/paperclip-tree, agent-team-foundation/first-tree`. Editing `~/.breeze/config.yaml` alone did not update the running daemon's allowlist — the allowlist is baked in at `start` time via `--allow-repo`. Required an explicit `stop` + `start --allow-repo <full csv>`.

Suggestions:
- `breeze status` could warn when `~/.breeze/config.yaml` `repos:` diverges from the daemon's `allowed repos` (say: "config has N repos, daemon allows M — restart to apply").
- Or: `breeze reload` that restarts with the config's current `repos:` list as the allowlist, so users don't have to re-type the csv.

### 3. `gardener-onboarding` skill doesn't mention the `--allow-repo` requirement
The skill's pull-mode section says just `first-tree breeze install && first-tree breeze start`. In practice both require `--allow-repo`. Updating the skill to show the real command would be useful.

### 4. Pull mode silently depends on GitHub watch subscription
Pull mode requires the user to be watching the source repo (otherwise no notifications → nothing to poll). The skill doesn't mention this, and `gh api /repos/<owner>/<repo>/subscription` requires the `notifications` scope that `gh auth login` doesn't grant by default.

Suggestions:
- `breeze doctor` (or a new `breeze check-watches`) could verify each repo in `repos:` has an active subscription and print a Watch URL when missing.
- `gardener-onboarding` skill should call this out explicitly with the Watch URL for the source repo.

### 5. `installed_version` in `.claude/gardener-config.yaml` drifts silently
[...]

PR #301, opened 2026-04-23
fix(breeze/start): refuse and warn when a live daemon exists

### LINKED ISSUE #293: breeze start: silent no-op when daemon already running with different --allow-repo
Split from #291 item 2.

## Problem
Daemon was running (pid 97184) with `allowed repos: A,B`. Editing `~/.breeze/config.yaml` alone did not update the live allowlist — the allowlist is baked in at `start` time. Required explicit `stop` + `start --allow-repo <full csv>`. Running `breeze start` a second time silently no-op'd instead of warning about the divergence.

## Fix ideas (pick one, or combine)
- \`breeze status\` warns when \`~/.breeze/config.yaml\` \`repos:\` diverges from the daemon's live \`allowed repos\`.
- \`breeze start\` errors (or prints a clear warning) if a daemon is already running and the requested allowlist differs.
- New \`breeze reload\` that restarts with \`config.yaml\` \`repos:\` as the allowlist, so users don't re-type the csv.

## Acceptance
- Running \`breeze start\` with a diverging config produces a visible message (not a silent no-op).
- One of the three reconciliation paths above is available.

Related: #294 (breeze add-repo ergonomic).

serenakeyitan:
## Summary

- \`breeze start\` now checks the service lock before bootstrapping launchd.
- If a non-stale lock exists (a daemon is already running), print the pid, explain that the live allow-list can't be updated from here, point at \`first-tree breeze stop\`, and exit 1.
- No behavior change when no daemon is running — launchd bootstrap path is unchanged.

## Motivation

Tester (#291 item 2) ran \`breeze start --allow-repo <new csv>\` while a daemon was already running with a different list. The command appeared to succeed but the live allow-list didn't update, because launchd's \`kickstart\` is idempotent on label and doesn't re-read args. This made it look like the allow-list edit took effect when it silently didn't.

This is the minimal fix: exit loudly when we detect a live daemon, so the user knows to \`stop\` first. The richer \"reconcile config.yaml \`repos:\` against the live allow-list automatically\" story is left to #297 (\`breeze add-repo\`).

## Refs

Closes #293
Umbrella: #291

## Test plan

- [x] \`tests/breeze/breeze-start.test.ts\` now has 3 tests: missing allow-repo, happy path (with \`findServiceLock\` mocked to null), and the new \"daemon already running\" path.
- [ ] Smoke test locally: start breeze, run \`first-tree breeze start --allow-repo ...\` again, confirm the error message appears and launchd is not re-kickstarted.

### DIFF
--- a/src/products/breeze/engine/commands/start.ts
+++ b/src/products/breeze/engine/commands/start.ts
@@ -95,6 +95,10 @@ export async function runStart(
     profile,
   );
   if (existingLock && !isLockStale(existingLock)) {
+    const stopCmd = formatStopCommand({
+      home: options.runnerHome ?? parseHome(argv),
+      profile: options.profile ?? parseProfile(argv),
+    });
     write(
       `breeze: daemon already running (pid ${existingLock.pid}).`,
     );
@@ -105,7 +109,7 @@ export async function runStart(
       "  will not update if you edit ~/.breeze/config.yaml or re-run `start`.",
     );
     write(
-      "  Run `first-tree breeze stop` first, then re-run `start` with the",
+      `  Run \`${stopCmd}\` first, then re-run \`start\` with the`,
     );
     write("  full --allow-repo csv.");
     return 1;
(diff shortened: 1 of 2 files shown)


bingran-you (changes requested):
Blocking issue:
- `src/products/breeze/engine/commands/start.ts:107`: the new refusal path always tells the user to run `first-tree breeze stop`, but both `start` and `stop` support `--home` and `--profile`. If the live daemon is under a non-default home/profile, that follow-up command targets the wrong runner and leaves the user stuck. Please either include the resolved flags in the message or make the guidance explicitly profile/home-aware.

bingran-you (approved):
Rechecked the follow-up commit. The stop guidance now carries through the resolved --home/--profile flags, which addresses the earlier blocker. Focused validation also passed (`pnpm typecheck`; `env -u BREEZE_DIR -u BREEZE_HOME pnpm exec vitest run tests/breeze/breeze-start.test.ts`). This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

PR #310, opened 2026-04-23
fix(gardener/sync): auto-seed sync-proposal label on tree repo (#303)

### LINKED ISSUE #303: Bug: gardener sync --open-issues silently drops the sync-proposal label when it doesn't exist
## Repro

Fresh tree repo without a pre-existing `first-tree:sync-proposal` label:

```
TREE_REPO_TOKEN=$(gh auth token) first-tree gardener sync --tree-path . --open-issues
```

## Observed

Each `gh issue create` emits:

```
⚠ label rejected on <tree-repo> (could not add label: 'first-tree:sync-proposal' not found) — retrying without --label
✒ opened issue on <tree-repo>: https://github.com/.../issues/NNN (assignees: ...)
```

Issues are created, but **without any label**. This silently breaks the downstream chain: breeze's draft-node dispatcher filters on `first-tree:sync-proposal`, so none of the filed issues match.

## Expected

One of:
1. Gardener auto-creates the label on the tree repo the first time it needs it.
2. Gardener hard-fails with a clear "create the label first" message.

Silent retry-without-label is the worst outcome — it looks like success but breaks the pipeline.

## Env

- first-tree main at 0.2.12.
- Discovered during onboarding test; 9 issues filed unlabeled before a separate rate-limit bug killed the run.

/cc @serenakeyitan

serenakeyitan:
Fixes #303.

## Summary

- Add `ensureTreeLabels()` in `sync.ts` that seeds `first-tree:sync-proposal`, `gardener`, and `needs-owner` on the tree repo once per `--open-issues` run, before any `gh issue create` call.
- "Already exists" on seed is treated as success. Other seed failures log a warning and fall through to the existing retry-without-label path — no regression on that degraded mode.
- Now issues keep their `first-tree:sync-proposal` label, so breeze's draft-node dispatcher actually sees them.

## Why this over the alternatives

The issue listed three options: (a) auto-create, (b) hard-fail, (c) keep silent-drop. Auto-create is idempotent, doesn't require operator prep, and fails gracefully — the hard-fail option would block first-time users of a fresh tree repo, and silent-drop is what caused the bug.

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run tests/gardener/sync-open-issues.test.ts` (20/20 pass, including 2 new #303 tests + the 18 pre-existing tests that verify the degraded retry-without-label path still works)
- [ ] E2E: re-run `first-tree gardener sync --open-issues` against a fresh tree repo without the labels; confirm issues are created with `first-tree:sync-proposal` label (deferred — will verify after merge during the next #291 e2e run)

### DIFF
--- a/src/products/gardener/engine/sync.ts
+++ b/src/products/gardener/engine/sync.ts
@@ -950,6 +950,77 @@ async function ghAuthOk(shellRun: ShellRun): Promise<boolean> {
   return result.code === 0;
 }
 
+/**
+ * Labels gardener applies to its sync-proposal issues. Kept in one place
+ * so `ensureTreeLabels` and the issue-create block agree.
+ */
+export const GARDENER_SYNC_LABELS: ReadonlyArray<{
+  name: string;
+  color: string;
+  description: string;
+}> = [
+  {
+    name: "first-tree:sync-proposal",
+    color: "0e8a16",
+    description: "Gardener-proposed tree update awaiting owner review",
+  },
+  {
+    name: "gardener",
+    color: "c5def5",
+    description: "Authored by the gardener automation",
+  },
+  {
+    name: "needs-owner",
+    color: "d93f0b",
+    description: "No NODE.md owner resolved — triage required",
+  },
+];
+
+/**
+ * Best-effort seed gardener's labels on the tree repo before we start
+ * opening issues. `gh label create` returns a 422 with "already exists"
+ * when the label is present — we treat that as success. Creation failures
+ * are non-fatal: the issue-create loop still has its legacy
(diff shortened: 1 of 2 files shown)


bingran-you (approved):
No blocking issues found. The new label-seeding path in runOpenIssuesMode addresses issue #303, and the added tests cover both first-run creation and the already-exists case. I also verified pnpm exec vitest run tests/gardener/sync-open-issues.test.ts and pnpm typecheck on the PR head. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

PR #312, opened 2026-04-23
fix(gardener/draft-node): strip trailing /NODE.md from marker (#306)

### LINKED ISSUE #306: Bug: gardener draft-node writes NODE.md inside a directory named NODE.md (doubled path)
## Repro

Run draft-node on a sync-proposal issue whose marker names a NODE.md target path:

```
TREE_REPO_TOKEN=$(gh auth token) first-tree gardener draft-node \
  --issue 318 --tree-repo serenakeyitan/paperclip-tree --tree-path .
```

Issue #318 body has marker:
```
<!-- gardener:sync-proposal · proposal_id=19aeb0ab73f4 · source_sha=unknown · node=product/task-system/stale-execution-lock-recovery/NODE.md -->
```

## Observed

Both dry-run and real run plan to write to:
```
product/task-system/stale-execution-lock-recovery/NODE.md/NODE.md
```

The real run opens a PR (serenakeyitan/paperclip-tree#336) that actually creates `NODE.md/` as a directory and writes a file named `NODE.md` inside it. The resulting path is not a valid tree node and is impossible to navigate via the normal NODE.md lookup.

## Expected

Path should be `product/task-system/stale-execution-lock-recovery/NODE.md` (a single file).

The marker already ends in `/NODE.md`; draft-node appears to be treating the marker's `node=<path>` as a directory and appending `NODE.md`, producing the doubled path.

## Fix suggestion

Strip a trailing `/NODE.md` from the marker's node path before re-appending. Or just use the marker path as-is without appending.

## Env

- first-tree main at 0.2.12.
- Tree repo: serenakeyitan/paperclip-tree, PR #336, issue #318.
- Reproducible on every draft-node call with a marker whose `node=` ends in `NODE.md` (which is all of them, since sync writes markers that way).

/cc @serenakeyitan

serenakeyitan:
Fixes #306.

## Summary

- Add `normalizeMarkerNode()` that strips an optional trailing `/NODE.md` (case-insensitive on the filename) plus any trailing slashes from the marker's `node=` field.
- Apply it inside `parseProposalMarker` so every consumer of `marker.node` (the path-append in `runDraftNode`, the dry-run output, the commit message, the PR title) sees the directory form.
- draft-node is the single authority for "where NODE.md lives inside a node directory" — collapsing the marker to the directory form is the right layer to fix this.

## Why parse-side (not write-side)

Per #306, the in-the-wild markers already have `node=.../NODE.md`, so any fix that only touched the write-side in `sync.ts` wouldn't repair paperclip-tree#318's existing issue and the others like it. A parse-side normalization catches both existing state and any future write-side drift (the classifier output is LLM-produced and has already shown this bug once).

The write-side in `buildSyncProposalBody` passes `proposal.path` straight through — if we want to prevent the bad form from being written at all, that's a separate follow-up that needs to touch every place `proposal.path` is used (issue title, body `Proposed tree update for …`, log lines, etc.) and decide what the canonical form is. Out of scope for this bug fix.

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run tests/gardener/draft-node.test.ts` — 28/28 pass, including 3 new normalization cases.
- [ ] E2E: re-dispatch draft-node against paperclip-tree#318 after merge, confirm the new PR writes `product/task-system/stale-execution-lock-recovery/NODE.md` (single file) instead of the doubled path.

### DIFF
--- a/src/products/gardener/engine/draft-node.ts
+++ b/src/products/gardener/engine/draft-node.ts
@@ -153,6 +153,36 @@ const MARKER_RE =
   /<!--\s*gardener:sync-proposal\s+([^>]+?)\s*-->/;
 
 /**
+ * Normalize the `node=<path>` field from a sync-proposal marker.
+ *
+ * Earlier gardener versions wrote the marker's `node=` as the tree
+ * directory path (`product/task-system/stale-execution-lock-recovery`),
+ * but a later change started emitting the trailing `/NODE.md`
+ * (`…/stale-execution-lock-recovery/NODE.md`). draft-node always appends
+ * `NODE.md` before writing, which silently produced the doubled
+ * `…/NODE.md/NODE.md` path seen on paperclip-tree#336.
+ *
+ * Treat both forms as equivalent: the marker names a node directory,
+ * and draft-node is the single authority on where `NODE.md` lives
+ * inside it. Strips optional trailing `/NODE.md` (case-insensitive on
+ * the filename). Also trims trailing slashes so `join()` behaves.
+ *
+ * Fixes #306.
+ */
+export function normalizeMarkerNode(raw: string | undefined): string {
+  if (!raw) return "";
+  let out = raw.trim();
+  // Strip a trailing `/NODE.md` (case-insensitive on the basename).
+  const withoutNodeMd = out.replace(/\/NODE\.md\/*$/iu, "");
+  if (withoutNodeMd !== out) {
+    out = withoutNodeMd;
+  }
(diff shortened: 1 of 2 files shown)


bingran-you (approved):
Reviewed the marker-path fix against the gardener sync contract and draft-node write path. The parse-side normalization is the right place to repair existing bad markers, and the added coverage hits the broken form plus the no-op cases. I did not find a blocking issue in this patch. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

Issue #305, opened 2026-04-23
Bug: gardener sync classification is non-deterministic; re-runs produce different proposals


serenakeyitan:
## Repro

Run `sync --open-issues --dry-run`, then immediately run `sync --open-issues` for real. Identical binding, identical tree state, identical source commit range.

## Observed

Dry-run: `58 total proposal(s) across 54 source PR(s)`.
Real run minutes later: `44 total proposal(s) across 54 source PR(s)`.

Same 54 source PRs, same tree, but 14 fewer proposals the second time. The divergence comes from `claude` CLI classifying the same drift differently on re-invocation (TREE_OK vs proposal) — expected LLM non-determinism, but problematic for a sync tool.

## Why this matters

1. **Retry isn't safe.** If a run partially fails (see rate-limit sibling bug), re-running can file a different set of issues than the original. Recovery is hard.
2. **Testing is unreliable.** Can't use dry-run to preview counts / assignees before committing, because the real run produces a different set.
3. **Multiple owners on the tree will see inconsistent proposal sets** if sync runs on different machines.

## Expected

Pick one:
1. **Cache classification results** keyed by (source_sha, tree_sha, prompt_version). Dry-run writes to cache; real run reads from cache.
2. **Pass temperature=0** (or equivalent) to the claude invocation.
3. **Multi-sample + vote** (classify each PR N times, take majority).

(1) is cheapest and gives true idempotence for the retry/rate-limit path.

## Env

- first-tree main at 0.2.12.
- Tree: serenakeyitan/paperclip-tree.

/cc @serenakeyitan

PR #314, opened 2026-04-23
fix(gardener): rate-limit backoff for sync --open-issues (#304)

### LINKED ISSUE #304: Bug: gardener sync --open-issues has no rate-limit backoff or resume-from-checkpoint
## Repro

Run `--open-issues` on a tree binding with many pending proposals (44 in my case):

```
TREE_REPO_TOKEN=$(gh auth token) first-tree gardener sync --tree-path . --open-issues
```

## Observed

Gardener bursts all `gh issue create` calls back-to-back. After ~10 issues it hit GitHub's GraphQL rate limit:

```
✒ opened issue on serenakeyitan/paperclip-tree: .../issues/326 (assignees: serenakeyitan)
❌ failed to open issue for proposal 12b5ef4c8edc: GraphQL: API rate limit already exceeded for user ID 94026305.
... (34 more identical failures) ...
✓ --open-issues paperclip-e392f6b1: 9 created, 0 skipped, 35 failed
```

No retry, no backoff, no resume. Only recourse is re-running the full command, which re-classifies all 54 source PRs via `claude` (expensive in tokens and minutes) and hits the same limit.

Compounding: re-runs can produce different proposals because the classifier is non-deterministic (filed separately as bug) — so re-running isn't even idempotent.

## Expected

1. Backoff on `GraphQL: API rate limit already exceeded` (sleep until reset).
2. Checkpoint already-created issues locally so re-runs skip proposals that already have an open tree issue.
3. Ideally: classify once, save results, then fire issue creation as a separate resumable phase.

## Env

- first-tree main at 0.2.12.
- 54 source PRs → 44 proposals → 9 filed, 35 lost.

/cc @serenakeyitan

serenakeyitan:
Fixes #304.

## Summary

- Add `runWithRateLimitBackoff()` + `isRateLimitError()` helpers to `sync.ts`.
- Wrap every `gh issue create` invocation in `runOpenIssuesMode` with the backoff helper.
- On a rate-limit stderr (covers GraphQL "API rate limit already exceeded", REST "rate limit exceeded", secondary-rate-limit, abuse-detection, and 429), sleep for exponentially growing intervals (30s → 60s → 120s → 240s, default 4 retries) and retry the same args.
- Non-rate-limit errors short-circuit — the existing 422 retry ladder (label-strip → assignee-strip → label-strip-again) is unchanged.

## Why within-run backoff

Reporter hit GitHub's per-user limit after ~10 of 44 creates and lost the other 35 because there was no retry. The existing `proposal_id` idempotency would have skipped them on re-run — but re-running re-classifies 54 source PRs through the `claude` CLI (expensive) and, per #305, isn't deterministic. Keeping the run alive across a single 60s limit window is the cheapest fix; within-run backoff complements the existing idempotency rather than replacing it.

Checkpoint-to-disk + classify-once-then-fire-as-resumable-phase from the issue description are separate, larger changes — this PR is the minimal bleed-stopper.

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run tests/gardener/sync-open-issues.test.ts` — 26/26 pass, including 6 new cases:
  - `isRateLimitError` recognises GraphQL/secondary/429 wordings, rejects 422 + empty
  - `runWithRateLimitBackoff` retries on rate-limit and succeeds
  - returns non-rate-limit errors immediately (no retry)
  - gives up after `maxRetries` on persistent rate-limit
- [ ] E2E: re-run the #304 reporter's 44-proposal scenario after merge and confirm 0 rate-limit losses.

### DIFF
--- a/src/products/gardener/engine/sync.ts
+++ b/src/products/gardener/engine/sync.ts
@@ -1526,6 +1526,54 @@ async function existingProposalIssueUrl(
   }
 }
 
+/**
+ * Detect GitHub rate-limit errors surfaced by `gh`. Covers both the
+ * REST "rate limit exceeded" wording and the GraphQL "API rate limit
+ * already exceeded" wording reported in #304.
+ */
+export function isRateLimitError(stderr: string): boolean {
+  if (!stderr) return false;
+  const s = stderr.toLowerCase();
+  return (
+    s.includes("rate limit") ||
+    s.includes("api rate limit") ||
+    s.includes("secondary rate limit") ||
+    s.includes("abuse detection") ||
+    s.includes("429")
+  );
+}
+
+/**
+ * Fixes #304: run a single `gh issue create` invocation with
+ * exponential backoff on GitHub rate-limit errors. A 44-proposal run
+ * was losing 35 of 44 issues to back-to-back 429s with no recovery;
+ * per-call backoff keeps the run alive long enough for the existing
+ * proposal_id idempotency to handle any true failures on re-run.
+ */
+export async function runWithRateLimitBackoff(
+  shellRun: ShellRun,
+  args: string[],
+  env: NodeJS.ProcessEnv,
+  opts: { maxRetries?: number; baseDelayMs?: number; sleep?: (ms: number) => Promise<void> } = {},
+): Promise<{ code: number; stdout: string; stderr: string }> {
+  const maxRetries = opts.maxRetries ?? 4;
+  const baseDelayMs = opts.baseDelayMs ?? 30_000;
+  const sleep =
+    opts.sleep ?? ((ms: number) => new Promise<void>((r) => setTimeout(r, ms)));
+  let res = await shellRun("gh", args, { env });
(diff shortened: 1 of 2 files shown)


bingran-you (approved):
Reviewed the new rate-limit helper and its integration into the existing label/assignee fallback ladder in `runOpenIssuesMode`; I don't see a blocking issue in the changed path. I also ran `pnpm typecheck` and `pnpm exec vitest run tests/gardener/sync-open-issues.test.ts` on this branch, and both passed locally.

yuezengwu (approved):
Reviewed the rate-limit backoff helper and its integration with the existing `--open-issues` issue-create fallback ladder. I did not find any blocking issues; non-rate-limit failures still flow back to the existing label/assignee retry paths, while rate-limit failures retry the same `gh issue create` args with exponential backoff.

Validated locally with `pnpm typecheck` and `pnpm exec vitest run tests/gardener/sync-open-issues.test.ts` (26 tests).

PR #315, opened 2026-04-23
fix(gardener): cache classifier results by prompt hash (#305)

### LINKED ISSUE #305: Bug: gardener sync classification is non-deterministic; re-runs produce different proposals

serenakeyitan:
Fixes #305.

## Summary

- Add a content-addressed on-disk cache at `.first-tree/classification-cache/<sha256(prompt)>.json`.
- Before each `claude` classifier call, hash the assembled prompt and check the cache. Hit → return cached proposals, skip the CLI entirely. Miss → run the classifier as before, persist on success, return.
- `FIRST_TREE_NO_CLASSIFIER_CACHE=1` opts out.
- `.first-tree/*` is already gitignored.

## Why hash the prompt (vs. the tuple)

The reporter recommended option 1: key by `(source_sha, tree_sha, prompt_version)`. The prompt we pass to `claude` already encodes all three:
- source commits + changed files + diff summary (per-PR content → source_sha in practice)
- tree-node paths/titles/owners + bodies for related nodes (tree_sha equivalent)
- the prompt template itself (prompt_version)

So sha256-of-prompt is one invalidation surface that captures all three, and the cache key is derivable without plumbing the dimensions separately. If any input changes — commits, diff, tree body, prompt template — the hash changes and we re-classify.

## What's NOT cached

- **Timeouts.** A 5-minute timeout is transient and should retry, not lock in the empty result.
- **Empty results.** Same rationale — an empty classification is a signal the CLI misbehaved, not a stable answer.
- **Cache write failures.** Best-effort; a read-only `.first-tree/` shouldn't break sync.

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run tests/gardener/classifier-cache.test.ts` — 6/6 pass (path layout, round-trip, malformed JSON, missing proposals array, versioned payload, miss returns null).
- [x] `pnpm exec vitest run tests/gardener/sync.test.ts tests/gardener/sync-open-issues.test.ts` — 49/49 pass (no regression from the added `treeRoot` parameter).
- [ ] E2E: re-run the reporter's 58-vs-44 scenario after merge — first run populates the cache, second run's proposal count matches exactly.

### DIFF
--- a/src/products/gardener/engine/sync.ts
+++ b/src/products/gardener/engine/sync.ts
@@ -721,12 +721,61 @@ function formatPrDiffForPrompt(
   return parts.join("\n");
 }
 
+/**
+ * Fixes #305. Classification results are cached on disk keyed by
+ * sha256(prompt) so re-runs (including `--dry-run` → real run) produce
+ * the same proposal set. The prompt already encodes source commits,
+ * tree-node state, changed files, and the prompt template itself, so
+ * hashing the prompt captures every input that should invalidate the
+ * cache. Set FIRST_TREE_NO_CLASSIFIER_CACHE=1 to bypass.
+ */
+export const CLASSIFIER_CACHE_DIR = ".first-tree/classification-cache";
+
+export function classifierCachePath(treeRoot: string, promptHash: string): string {
+  return join(treeRoot, CLASSIFIER_CACHE_DIR, `${promptHash}.json`);
+}
+
+export function readClassifierCache(
+  treeRoot: string,
+  promptHash: string,
+): ClassificationItem[] | null {
+  const path = classifierCachePath(treeRoot, promptHash);
+  if (!existsSync(path)) return null;
+  try {
+    const parsed = JSON.parse(readFileSync(path, "utf8")) as unknown;
+    if (!parsed || typeof parsed !== "object") return null;
+    const entry = parsed as { proposals?: unknown };
+    if (!Array.isArray(entry.proposals)) return null;
+    return entry.proposals as ClassificationItem[];
+  } catch {
+    return null;
+  }
+}
+
+export function writeClassifierCache(
+  treeRoot: string,
+  promptHash: string,
+  proposals: ClassificationItem[],
+): void {
+  const path = classifierCachePath(treeRoot, promptHash);
+  mkdirSync(dirname(path), { recursive: true });
+  writeFileSync(
+    path,
+    JSON.stringify(
+      { version: 1, promptHash, createdAt: new Date().toISOString(), proposals },
+      null,
+      2,
+    ),
+  );
+}
+
 async function classifyDriftViaClaude(
   shellRun: ShellRun,
   drift: DriftReport,
   treeNodes: TreeNodeSummary[],
-  changedFiles?: string[],
-  diffSummary?: string,
+  changedFiles: string[] | undefined,
+  diffSummary: string | undefined,
+  treeRoot: string,
 ): Promise<ClassificationItem[]> {
   // Build keywords from the PR for relevance matching
   const driftText = [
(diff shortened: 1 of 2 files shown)


bingran-you (approved):
## Review summary

Approving. This is the right fix for #305 and directly implements option 1 from the issue. Code is clean, tests cover the essentials, and the scope is tight.

## What I like

- **Hashing the full assembled prompt** is cleaner than plumbing `(source_sha, tree_sha, prompt_version)` separately — any change to the inputs or the template invalidates the cache automatically, with no version-bump discipline required.
- **What's not cached is correct:** timeouts, empty results, and cache-write failures all fail open to re-classification. That matches the spirit of the bug (we want re-runs to converge on a real answer, not lock in a transient bad one).
- `persist()` guards on `proposals.length > 0`, so empty arrays don't pollute the cache.
- Test file exercises the load-bearing paths: round-trip, malformed JSON, missing `proposals`, versioned payload.

## Non-blocking notes

1. **Cache-hit / bypass behavior isn't covered by a test.** The core read/write behavior is tested, but `FIRST_TREE_NO_CLASSIFIER_CACHE=1` and the "cache-hit skips the classifier call" guarantee are only verified by reading the code. A test that stubs `shellRun` and asserts it isn't called on a cache hit would lock the main invariant in.

2. **`version: 1` is written but never checked on read.** Today the prompt-hash also isolates schema drift (if the payload shape changes alongside a prompt/template change, the hash changes too). But if you ever change `ClassificationItem`'s shape without touching the prompt, stale caches would silently deserialize into the new shape. Cheap to add a `if (entry.version !== 1) return null;` guard now and save a future head-scratch.

3. **Prompt determinism depends on caller-side ordering.** The prompt composition here is deterministic given the inputs, but `treeNodes`, `drift.commits`, `drift.mergedPrTitles`, and `changedFiles` are consumed in whatever order they arrive. If any of those upstream sources aren't ordered deterministically (e.g. filesystem iteration, API pagination), two runs on identical state could still hash differently. Worth a follow-up pass on the callers to confirm, but not a blocker for this PR.

4. **Cache lives in the source workspace checkout (`treeRoot`).** That means dry-run → real-run idempotence holds within one workspace, which is what the issue asked for. Cross-machine / cross-CI-run idempotence (the third item under "Why this matters") isn't addressed here — that would need either a shared cache or a deterministic classifier. Worth documenting explicitly if users expect the stronger guarantee.

## Tree-sync CI failure

The failing `tree-sync` check on this PR is the standard forked-branch auth issue (`TREE_REPO_TOKEN` unavailable to the workflow), not related to this change. Code CI is green.

---

PR #316, opened 2026-04-23
fix(#296): drop unused installed_version field from gardener-config

### LINKED ISSUE #296: gardener-config: installed_version field drifts silently
Split from #291 item 5.

## Problem
Existing \`.claude/gardener-config.yaml\` had \`installed_version: v2.4.1\`. Latest CLI reports \`first-tree=0.2.10 gardener=0.4.0\`. The two aren't on the same version scheme, so the field is clearly stale from an older setup with no mechanism to refresh.

## Fix options
- Remove the field entirely (treat CLI version as single source of truth).
- Auto-update on every gardener run (write current CLI version back).
- \`breeze doctor\` flags drift without touching the file.

## Acceptance
- Decision made on which option.
- Either the field is gone, auto-refreshed, or surfaced by doctor.

serenakeyitan:
Fixes #296.

## Summary

- Drop `installed_version` from the `GardenerConfig` type, parser, and docblock example.
- No call-site reads it; the field had no consumer. Operators couldn't tell if it was stale because nothing ever rewrote it (the reported drift: config said `v2.4.1`, CLI reports `first-tree=0.2.10 gardener=0.4.0` — different scheme entirely).
- Parser still *accepts* a legacy `installed_version:` line in existing tree configs — it's just silently ignored, so nobody with that line in a checked-in config breaks.

## Why Option A

Of the three options in the issue body:
- **A. Remove the field** ← this PR. Zero drift possible. Simplest.
- B. Auto-update on every run. Rejected: mutates config on read-only commands (including `--dry-run`), creates surprising git diffs, races between concurrent gardeners.
- C. `breeze doctor` flags drift. Rejected: `breeze doctor` doesn't exist yet, and in practice the field stays stale because nobody runs doctor.

The audit-trail use case ("which CLI version installed this config?") is hypothetical. If we ever need it we can reintroduce a read-only field keyed off git history of the config file.

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run tests/gardener/gardener-respond.test.ts` — 41/41 pass.
- [x] `pnpm exec vitest run` — 1188/1188 pass, no regressions.

### DIFF
--- a/src/products/gardener/engine/runtime/config.ts
+++ b/src/products/gardener/engine/runtime/config.ts
@@ -8,7 +8,6 @@
  *
  * Supported shape:
  *   tree_repo: owner/name
- *   installed_version: 0.1.0
  *   target_repos:
  *     - owner/repo-one
  *     - owner/repo-two
@@ -33,7 +32,6 @@ export interface ModuleToggle {
 
 export interface GardenerConfig {
   tree_repo?: string;
-  installed_version?: string;
   target_repos?: string[];
   modules?: {
     sync?: ModuleToggle;
@@ -209,9 +207,6 @@ function coerceModuleToggle(value: unknown): ModuleToggle | undefined {
 function coerceConfig(raw: Record<string, unknown>): GardenerConfig {
   const config: GardenerConfig = {};
   if (typeof raw.tree_repo === "string") config.tree_repo = raw.tree_repo;
-  if (typeof raw.installed_version === "string") {
-    config.installed_version = raw.installed_version;
-  }
   const targets = coerceStringArray(raw.target_repos);
   if (targets) config.target_repos = targets;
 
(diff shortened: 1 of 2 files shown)


bingran-you (approved):
LGTM. Option A is the right call for the reasons you laid out — no consumer, no refresh mechanism, no drift possible once it's gone. The "silently accept and ignore legacy line" behavior falls out naturally from `coerceConfig` dropping unknown keys, and the regression test in `tests/gardener/gardener-respond.test.ts` nails that contract so nobody accidentally tightens it into an error later.

Verified locally:
- `rg installed_version` on the branch returns zero production hits (only the new regression test).
- Docblock in `src/products/gardener/engine/runtime/config.ts` is updated to match the shrunken shape.
- CI is green; the `tree-sync` failure is unrelated to this change.

Nit (non-blocking): if you ever want the audit trail back, keying it off `git log -- .claude/gardener-config.yaml` as you suggested in the description is strictly better than a self-reported field. No action needed here.

PR #322, opened 2026-04-23
tree: queue GitHub native auto-merge after opening tree PRs

serenakeyitan:
## Summary

- \`openTreePr\` now calls \`gh pr merge <url> --auto --squash --delete-branch\` after successful PR creation, so review-approved tree PRs merge themselves. This closes the end-to-end loop promised by the onboarding (\"source PR → tree issue → breeze pickup → merged tree PR\").
- Per-tree control lives entirely on the target repo: GitHub only honors \`--auto\` when the repo has \"Allow auto-merge\" enabled in Settings → General. Every repo that hasn't opted in (the default, including this one) gets a silent non-zero exit from \`gh\` and sees no behavior change.
- Skip queueing when the PR already existed (\"a pull request for branch already exists\" path) — we don't re-touch merge state on branches we didn't create this run.

## Why not a YAML/daemon-level flag

Considered a \`auto_merge:\` key in \`.claude/gardener-config.yaml\` with per-tree gates (\`allowed_authors\`, \`max_files_changed\`, etc.). Dropped it: too much config surface for an OSS tool, and it duplicates controls GitHub already provides (required reviews, required checks, branch protection). One switch, one place, auditable in the tree repo's Settings.

## Test plan

- [x] \`pnpm typecheck\` clean
- [x] \`pnpm test\` — 1191 passed / 51 skipped. \`tests/tree/open-tree-pr.test.ts\` gains three new cases: asserts the \`--auto --squash --delete-branch\` argv, skips the call on the \"already exists\" path, and tolerates non-zero exit from \`gh pr merge\` (repo with auto-merge disabled).
- [x] \`tests/fixtures/sync-golden/two-pr-apply.json\` regenerated via \`UPDATE_SYNC_GOLDEN=1\` — shell-call envelope repo-gardener parses now includes the \`gh pr merge\` lines.
- [ ] Behavior check on paperclip-tree after merge: owner enables \"Allow auto-merge\" in Settings → General → next breeze-opened PR carries the queued auto-merge and lands when bingran approves.

Closes #321.

PR #318, opened 2026-04-23
docs(readme): add end-to-end gardener → breeze onboarding prompt

serenakeyitan:
## Summary

Adds a third paste-into-Claude-Code block under README's **Quick Start For Agents** for users who want to drive the full drift cycle end-to-end: source PR → tree issue → breeze pickup → draft-node PR.

## The new block

> Install the first-tree CLI from [releases](https://github.com/agent-team-foundation/first-tree/releases), then run \`gardener install\` to bind a tree to my source repo and \`breeze install\` to route notifications back to me. Trigger one drift event end-to-end: source PR → tree issue → breeze pickup → draft-node PR.
>
> Monitor: [paperclip](https://github.com/paperclipai/paperclip) (source) and [paperclip-tree](https://github.com/serenakeyitan/paperclip-tree) (tree).

## Test plan

- [x] README renders cleanly (markdown lint clean).
- [x] No changes outside `README.md`.

This reply was authored by Claude Code on behalf of the account owner.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #317, opened 2026-04-23
release: 0.2.13

serenakeyitan:
## Summary

Bump CLI 0.2.12 → 0.2.13 (gardener product 0.4.0 → 0.4.1).

## Gardener fixes shipped today

| Issue | Fix | PR |
|---|---|---|
| #307 | `gardener respond` accepts `first-tree/draft-node-*` branches | #311 |
| #306 | draft-node strips trailing `/NODE.md` from marker (no more `foo/NODE.md/NODE.md`) | #312 |
| #302 | `--assignee USER` override for `sync --open-issues` (with CLI plumbing regression test) | #313 |
| #304 | exponential rate-limit backoff for `sync --open-issues` | #314 |
| #305 | content-addressed classifier cache keyed by sha256(prompt) | #315 |
| #296 | drop unused `installed_version` field from gardener-config (legacy line tolerated) | #316 |

## Compatibility

- No breaking changes. Parser still silently accepts a legacy `installed_version:` line in existing tree configs.
- `classifier-cache/` directory is created on demand under `.first-tree/` (already gitignored). Opt out with `FIRST_TREE_NO_CLASSIFIER_CACHE=1`.

## Release check

- [x] `pnpm version:check` — CLI 0.2.13, product 0.2.13, skill 0.2
- [x] `pnpm validate:skill`
- [x] `pnpm typecheck`
- [x] `pnpm test` — 1188/1188 pass
- [x] `pnpm build`
- [x] `pnpm test:dist` — 12/12
- [x] `pnpm test:release` — 19/19 (pack-contents + install-smoke)

## Post-merge

1. Pull `main`, tag `v0.2.13`, push tag.
2. `npm publish` from a clean checkout of the tagged commit.
3. Create GitHub Release with the same body.

This reply was authored by Claude Code on behalf of the account owner.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #275, opened 2026-04-22
chore(release): 0.2.12

serenakeyitan:
## Summary
Version bump 0.2.11 → 0.2.12. Ships the #272 fix for #269.

## What's included
- feat(gardener/classifiers): add claude-cli classifier with auth-failure fallback (#272) — closes #269
  - Shells out to local `claude` so OAuth-authenticated Claude Code users no longer need `ANTHROPIC_API_KEY`
  - `GARDENER_CLASSIFIER` env var escape hatch for CI/testing (`claude-cli` | `anthropic-api` | `none`)
  - Transparent fallback to `anthropic-api` if claude-cli throws auth failure AND the key is set

## Test plan
- [x] `pnpm version:check` reports "Versions match: CLI 0.2.12, product 0.2.12, skill 0.2"
- [ ] After merge: tag `v0.2.12`, `npm publish` (manual)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #323, opened 2026-04-23
release: 0.2.14

serenakeyitan:
## Summary

Bumps first-tree 0.2.13 → 0.2.14 and gardener 0.4.1 → 0.4.2. Picks up the auto-merge queueing in \`openTreePr\` and the onboarding-prompt fix that shipped to \`main\` today.

## What's in 0.2.14

- **#322** tree: queue GitHub native auto-merge after opening tree PRs. \`openTreePr\` now calls \`gh pr merge --auto --squash --delete-branch\` after PR create. Per-repo opt-in via the target repo's \"Allow auto-merge\" Settings toggle; repos that haven't opted in get a silent swallow of the specific disabled-on-repo stderr. Auth / transport / unexpected failures now surface as real errors instead of being swallowed.
- **#318** docs(readme): fix end-to-end gardener → breeze onboarding prompt. Replaces the nonexistent \`gardener install\` with \`first-tree tree inspect\` → bind → \`gardener install-workflow\` → \`breeze install\`, and flags the \`ANTHROPIC_API_KEY\` + \`TREE_REPO_TOKEN\` secrets prerequisite.
- **#316** fix: drop unused \`installed_version\` field from gardener-config.

## Release check

Local \`pnpm release:check\` pipeline (version sync → skill validate → typecheck → 1192 tests → build → 12 dist tests → 19 release tests) all green.

## Test plan

- [ ] CI green on this PR
- [ ] After merge: tag \`v0.2.14\`, create GitHub Release (no npm publish per owner preference)


yuezengwu (changes requested):
I found two release-contract issues in this version-bump PR:

1. `assets/tree/VERSION` is being bumped even though no shipped tree runtime assets changed since `release: 0.2.13`. Per `docs/architecture/versioning.md`, that version should track installable asset payload changes, not package releases. Bumping it here would trigger unnecessary asset overwrites during `tree upgrade`.

2. `src/products/tree/VERSION` is being bumped even though the only behavior change since `release: 0.2.13` is the auto-merge path in `openTreePr`, which is called from `gardener` (`src/products/gardener/engine/sync.ts` and `draft-node.ts`). I couldn't find a `tree` CLI command that uses that primitive, so this looks like a gardener surface change rather than a tree surface change.

`package.json` and `src/products/gardener/VERSION` make sense for this release.

serenakeyitan:
Good catches on the substance, but there's a gate conflict we need to resolve before I can act on the revert.

`scripts/sync-version.js` (the first stage of `release:check`) **requires** `package.json`, `assets/tree/VERSION`, and `src/products/tree/VERSION` to be byte-identical. Quoting lines 7/8 of that script:

> 1. `package.json` `version` — full major.minor.patch (CLI)
> 2. `assets/tree/VERSION` — full major.minor.patch (tree product)
> 4. `src/products/tree/VERSION` — full major.minor.patch (tree product)
> The first two and the fourth must be identical.

So reverting `assets/tree/VERSION` and `src/products/tree/VERSION` to `0.2.13` while keeping `package.json` at `0.2.14` makes `pnpm release:check` fail at the first stage. That's the contract I was following here (and that every prior release PR followed — #317, #275, #273 all bump all four).

So there are **two load-bearing contracts in conflict**:

- `docs/architecture/versioning.md`: tree VERSION files bump only on tree-asset / tree-CLI surface changes (your read — correct for this PR).
- `scripts/sync-version.js`: tree VERSION files must equal `package.json` version always.

Resolving that conflict is bigger than this release PR. Two options, happy to take whichever you prefer:

**Option A — ship 0.2.14 as-is** (tree VERSIONs move in lockstep with package.json, per today's gate). Document the `sync-version.js` ↔ `versioning.md` conflict as a follow-up issue. Low risk of asset-overwrite churn because `tree upgrade` also checks content hashes, not just the VERSION file — but I'd want to verify that claim before relying on it.

**Option B — fix the contract first**. Open a separate PR that either (a) relaxes `sync-version.js` to allow independent tree VERSIONs, or (b) updates `versioning.md` to say tree VERSIONs track package version. Then come back to this release.

Which direction?

serenakeyitan:
Shipping 0.2.14 as-is. Reasoning:

1. Every prior release (#273, #275, #317) bumped all four files in lockstep — this PR matches precedent.
2. `scripts/sync-version.js` is enforced code; `versioning.md` is docs that nobody's enforced. When they disagree, the runtime gate is the real contract.
3. The tester is waiting on the 0.2.14 fixes (`openTreePr` auto-merge + onboarding prompt). A doc-vs-script reconciliation shouldn't block that.

Filed the contract conflict as a separate tracking issue so we decide which rule wins without gating this release on it.

yuezengwu (approved):
Thanks for filing #324. I checked the current code path and I'm okay unblocking 0.2.14 on the basis of the enforced `version:check` gate plus the follow-up issue, so I'm dropping the release blocker for this PR.

One factual correction for the tracking thread: `tree upgrade` does not currently gate on asset content hashes. The bundled upgrade path compares the installed skill version at major.minor granularity and, when it does run, `refreshShippedWorkflows` overwrites shipped workflow files unconditionally. So I agree with shipping this PR as precedent-preserving, but I don't want the content-hash claim to become the accepted explanation for why it's safe.

PR #325, opened 2026-04-23
docs: make end-to-end onboarding prompt source/tree-agnostic

serenakeyitan:
## Summary

The end-to-end onboarding prompt in README.md hardcoded `paperclip` (source) and `paperclip-tree` (tree), so pasting it as-is pointed every new user at a demo repo they don't own. Restructures the prompt so the user names their **source repo** and **tree repo** as explicit fields at the top, and drops the paperclip monitor URLs.

Also removes a stray `gardener install` mention that referenced a command we don't ship (the real command is `gardener install-workflow`).

## Test plan

- [x] Re-read the onboarding section end to end — prompt is now copy-paste usable for any source+tree pair
- [x] Only command references in the block are real CLI commands (`tree inspect`, `gardener install-workflow`, `breeze install`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,25 +112,30 @@ https://github.com/<your-org>/<your-tree-repo>.
 
 **Driving one full gardener → breeze cycle end-to-end:**
 
-Pick the source repo (the codebase you ship from) and the tree repo (where
-`NODE.md` lives) before pasting. The agent asks for both up front and won't
-assume either.
+Paste this as-is — the agent asks you for the source repo and tree repo
+before running anything, so there's nothing to fill in up front.
 
 ```text
 Use the latest first-tree CLI (https://github.com/agent-team-foundation/first-tree).
 
-Source repo: <owner/source-repo>
-Tree repo:   <owner/tree-repo>
+Before running any command, ask me two questions and wait for my answers:
 
-Run `first-tree tree inspect --json` in the source repo, install the skill, and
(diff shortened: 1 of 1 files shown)


yuezengwu (approved):
LGTM — docs-only change and a net improvement over the previous copy.

- Swapping the hardcoded `paperclip` / `paperclip-tree` links for an explicit "ask the user for source + tree repos" turn in the prompt is the right call. Users copy-pasting this as-is previously would have pointed agents at repos they don't own; now it fails closed on missing/ambiguous answers instead of silently proceeding.
- The intro hint ("Paste this as-is — the agent asks you…") is a nice touch because the `source repo`/`tree repo` phrasing inside the fenced block could otherwise read like two placeholders the human is expected to fill in.
- Only real CLI commands referenced (`tree inspect`, `gardener install-workflow`, `breeze install`), consistent with what ships.

Minor, non-blocking nits (feel free to skip):
- "Monitor both repos I gave you." reads a little terse once the paperclip monitor line is gone — consider "Monitor the source repo and tree repo you gave me." for symmetry with the rest of the agent-voice wording. Totally optional.
- PR description mentions removing a stray `gardener install` reference, but I only see `install-workflow` being tightened up in the diff — worth double-checking there isn't another mention elsewhere in the README you meant to catch, but not a blocker for this PR.

Approving. CI (`ci`) is still running at the time of review; the `tree-sync` failure looks unrelated to this README-only change, but please eyeball it before merging.

PR #326, opened 2026-04-23
docs(onboarding): default to breeze-first setup; drop push/pull framing

serenakeyitan:
## Summary

- README: rewrite the "end-to-end" quick-start prompt to reference the onboarding skill and stay short. Previous version recommended `gardener install-workflow` (GitHub Actions) as the default, which requires source-repo write + `ANTHROPIC_API_KEY` + `TREE_REPO_TOKEN` + a workflow-approval PR — friction a first-time user doesn't need.
- `skills/first-tree/references/onboarding.md` Step 6: rewrite from "optional modules toggle" into agent-led setup for gardener + breeze. Two questions (source repo, watch status), `breeze install`, verify end-to-end. Workflow-mode deployment moved to a mention in Pitfalls for the niche power-user case; still reachable via `workflow-mode.md`.

## Why

Onboarding testing against v0.2.14 surfaced that the `install-workflow` path isn't realistic for most first-time users — breeze works for everyone (owners and non-owners), needs only `gh auth`, and is the simpler recommendation. Presenting workflow mode as a peer of breeze in the main onboarding flow implies a deployment choice newcomers don't need to make, and pulls them into PAT/secret setup before they've even seen the chain work.

Followup discussion from #291. The workflow path remains documented in `workflow-mode.md` and reachable via the Pitfalls section, so the niche case is still supported — it's just not the default.

## Test plan

- [ ] Read README § "Quick Start For Agents" — the third prompt now says "follow the onboarding skill" instead of walking the user through `install-workflow` + secrets inline.
- [ ] Read `skills/first-tree/references/onboarding.md` Step 6 — verify the flow is: ask 2 questions → verify watch → `breeze install` → trigger end-to-end → report each step. No "push mode" / "pull mode" terminology.
- [ ] Confirm `workflow-mode.md` is still discoverable (linked from the Pitfalls bullet in Step 6).

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,41 +112,24 @@ https://github.com/<your-org>/<your-tree-repo>.
 
 **Driving one full gardener → breeze cycle end-to-end:**
 
-Paste this as-is — the agent asks you for the source repo and tree repo
-before running anything, so there's nothing to fill in up front.
+Paste this as-is — the agent will ask you which repos to target before
+running anything, so there's nothing to fill in up front.
 
 ```text
-Use the latest first-tree CLI (https://github.com/agent-team-foundation/first-tree).
+I want to set up first-tree — a Context Tree that tracks decisions across
+my source repos. It ships two agents: gardener (detects drift and drafts
+tree updates) and breeze (a local daemon that dispatches gardener on
+GitHub notifications).
 
-Before running any command, ask me two questions and wait for my answers:
+Latest release: https://github.com/agent-team-foundation/first-tree/releases/latest
 
-1. "What is the source repo you ship from?" (owner/name on GitHub)
-2. "What is the tree repo that stores NODE.md?" (owner/name on GitHub)
-
-Do not assume, guess, or infer either from the current working directory.
-If either answer is missing or ambiguous, stop and re-ask. Only after I've
-confirmed both, proceed.
-
-Prerequisites: `gh` CLI authed (`gh auth status`), `jq` installed,
-Node ≥ 22, and `ANTHROPIC_API_KEY` exported locally.
-
-Then, using those two values:
-
-1. Clone both repos locally if they aren't already, and `cd` into the
-   source repo checkout before running anything else.
-2. Run `first-tree tree inspect --json` in the source repo to confirm
-   classification, then run `first-tree skill install` to install the
-   first-tree skill, and bind to the tree repo with
-   `first-tree tree bind --tree-path <path-to-tree-repo-checkout>`.
-3. From inside the source repo checkout, install the gardener push-mode
-   workflow: `first-tree gardener install-workflow --tree-repo <tree-repo>`.
-   Walk me through the required `ANTHROPIC_API_KEY` and `TREE_REPO_TOKEN`
-   secrets.
-4. Start breeze so notifications route back to me:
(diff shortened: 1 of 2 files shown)


serenakeyitan:
Rebased onto main after #325 merged. Resolved the README conflict by keeping this PR's shorter breeze-first prompt (right design — moves heavy `install-workflow` guidance out of the main prompt) and folding in the few bits from #325 that were additive:

- Added an explicit "ask me which source repo and which tree repo to target, and wait for my answers — do not infer from the current working directory" instruction inside the README prompt, so a cold agent doesn't silently skip the clarifying questions.
- Added a Prerequisites block at the top of Step 6 in `onboarding.md` (gh authed, jq, Node ≥ 22, `ANTHROPIC_API_KEY`) — onboarding E2E showed these were implicit and a paste-and-run user could get stuck without them.
- Step 6.1 now asks for source repo **and** tree repo explicitly (it was only asking for source), and states "do not infer from cwd."

Net: one commit on top of main, no new commits added (rebased into the existing commit).

README.md
```diff
@@ -112,41 +112,24 @@ https://github.com/<your-org>/<your-tree-repo>.
...
-   `gh pr list` / `gh issue list` on each.
+I want to set up first-tree — a Context Tree that tracks decisions across
+my source repos. It ships two agents: gardener (detects drift and drafts
+tree updates) and breeze (a local daemon that dispatches gardener on
+GitHub notifications).
+
+Latest release: https://github.com/agent-team-foundation/first-tree/releases/latest
+
+Install the CLI, then follow the onboarding skill at
+`.agents/skills/first-tree/references/onboarding.md`. Before running any
```
yuezengwu, line 127: This regresses the cold-start flow: after only installing the CLI, `.agents/skills/first-tree/references/onboarding.md` is not on disk yet. A literal agent following this prompt will fail before it ever reaches the step that installs the skills. Can we either restore an explicit `first-tree skill install` step before this reference, or send the agent to a CLI entrypoint that exists immediately after install (for example `first-tree tree help onboarding`)?

yuezengwu (changes requested):
One blocking issue: the new README quick-start prompt now tells a fresh agent to read `.agents/skills/first-tree/references/onboarding.md` immediately after CLI install, but that path only exists after the local skill payloads have been installed. That makes the literal-paste flow fail for the exact cold-start case this section is trying to optimize. Please restore an installable entrypoint before that path reference, or point the prompt at a CLI surface that already exists after install (for example `first-tree tree help onboarding`).

serenakeyitan:
Missed this by 6 seconds — review landed just after I admin-merged. The critique is correct: `.agents/skills/first-tree/references/onboarding.md` doesn't exist until `first-tree skill install` runs, which is exactly the chicken-and-egg the prompt is supposed to sidestep. Will land a follow-up PR that points the prompt at `first-tree tree help onboarding` (available immediately post-CLI-install) and moves the skill-path reference to after the skill install step.

yuezengwu:
Acknowledged. Since #326 is already merged, #327 is the right place to address the cold-start regression. No further action needed on this thread from my side; I'll review the fix there. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

PR #333, opened 2026-04-23
docs(onboarding): gate Step 6.2 watch check on notifications token scope

### LINKED ISSUE #332: Onboarding Step 6.2: `gh api /subscription` check fails with 404 without notifications token scope
## Where

`skills/first-tree/references/onboarding.md` Step 6.2 (from PR #326, merged into v0.2.15):

```
gh api /repos/<source-repo>/subscription
# 200 → watching
# 404 → not watching
```

## Problem

On a default `gh auth login` setup, the user's token does **not** include the `notifications` OAuth scope. With default scopes (e.g. `admin:org, gist, repo, workflow`), `/subscription` returns **404 for every repo**, whether watching or not. The agent following this doc will:

1. Run the check → get 404.
2. Conclude "not watching," tell the user to click Watch → All Activity.
3. User already is watching, clicks again, refreshes, reruns → still 404.
4. Stuck.

Repro on a token lacking `notifications`:
```
$ gh auth status 2>&1 | grep scope
Token scopes: 'admin:org', 'gist', 'repo', 'workflow'
$ gh api /repos/paperclipai/paperclip/subscription
{"message":"Not Found","status":"404"}
$ # user IS watching the repo — verified via the GitHub UI
```

`gh` itself tells you: *"This API operation needs the `notifications` scope. To request it, run: `gh auth refresh -h github.com -s notifications`"* — but that's only when `gh` can see it's being called without scope. From a raw `gh api` call, it just returns 404.

## Suggested fix

Update Step 6.2 to:

1. Check scope first: `gh auth status 2>&1 | grep -i 'notifications'`. If absent, instruct the user to run `gh auth refresh -h github.com -s notifications` (interactive) and re-run the check.
2. Only treat 404 as "not watching" after confirming the token has `notifications` scope.

Alternative: skip the programmatic watch check entirely and just ask the user to visit `https://github.com/<source-repo>` and confirm they see "Unwatch" rather than "Watch" on the repo page. That's low-tech but avoids the scope hurdle, and ~95% of users will be watching (or easily verifiable visually).

## Env

- Discovered during v0.2.15 onboarding test (2026-04-23).
- Tree: serenakeyitan/paperclip-tree, source: paperclipai/paperclip.
- Token: default `gh auth login` scopes.

/cc @serenakeyitan

serenakeyitan:
Fixes #332.

## Problem

`gh api /repos/<repo>/subscription` returns 404 for every repo when the user's `gh` token lacks the `notifications` OAuth scope — which default `gh auth login` does not grant (typical scopes are `admin:org, gist, repo, workflow`). An agent following Step 6.2 on a default token would:

1. Run the check → get 404.
2. Conclude "not watching," tell the user to click Watch → All Activity.
3. User already is watching, clicks again, refreshes, agent re-runs → still 404.
4. Stuck.

## Change

Step 6.2 now:

1. Runs `gh auth status | grep notifications` first.
2. If absent, tells the user to run `gh auth refresh -h github.com -s notifications` (interactive) and waits for confirmation.
3. Only then runs the `/subscription` check.

Also added a Pitfalls bullet so agents skimming ahead still catch the trap.

## Test plan

- [ ] Read Step 6.2: flow is scope check → refresh if needed → subscription check.
- [ ] Read Pitfalls bullet: explains 404-without-scope explicitly.
- [ ] No changes outside `skills/first-tree/references/onboarding.md`.

Discovered during v0.2.15 onboarding test.

### DIFF
--- a/skills/first-tree/references/onboarding.md
+++ b/skills/first-tree/references/onboarding.md
@@ -214,12 +214,15 @@ answer is missing or ambiguous, stop and re-ask.
 
 ### 6.2 Verify Watch Status
 
-First, confirm the user's `gh` token has the `notifications` scope —
-without it `/subscription` returns 404 for every repo, whether
-watching or not:
+First, confirm the user's active `github.com` `gh` token has the
+`notifications` scope — without it `/subscription` returns 404 for
+every repo, whether watching or not. Scope the check to the active
+`github.com` account; a bare `gh auth status | grep notifications`
+can false-positive on multi-host or multi-account setups where
+another stored account has the scope:
 
 ```bash
-gh auth status 2>&1 | grep -i 'notifications'
+gh auth status --active --hostname github.com 2>&1 | grep -i 'notifications'
 ```
 
 If the scope is missing, instruct the user to run:
@@ -375,12 +378,15 @@ ignore them.
   first-time setup).
 - Pull-based dispatch silently no-ops without a GitHub watch
   subscription. Verify before claiming setup is complete.
-- `gh api /repos/<repo>/subscription` returns 404 when the user's
-  token lacks the `notifications` scope — which default `gh auth
-  login` does **not** grant. Check `gh auth status` for the scope
-  first; if missing, run `gh auth refresh -h github.com -s
-  notifications`. Treat 404 as "not watching" only after the scope
-  is confirmed present.
+- `gh api /repos/<repo>/subscription` returns 404 when the active
+  `github.com` token lacks the `notifications` scope — which default
+  `gh auth login` does **not** grant. Check scope with
+  `gh auth status --active --hostname github.com 2>&1 | grep -i
+  notifications` (scope the check to the active account — a bare
+  `gh auth status` grep can false-positive on multi-host or
+  multi-account setups). If missing, run
+  `gh auth refresh -h github.com -s notifications`. Treat 404 as
+  "not watching" only after the scope is confirmed present.
 - `TREE_REPO_TOKEN` must be in the environment for every `gardener`
   invocation, not just at setup.
 
--- a/skills/first-tree/references/onboarding.md
+++ b/skills/first-tree/references/onboarding.md
@@ -214,6 +214,25 @@ answer is missing or ambiguous, stop and re-ask.
 
 ### 6.2 Verify Watch Status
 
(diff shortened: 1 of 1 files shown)


skills/first-tree/references/onboarding.md
```diff
@@ -214,6 +214,25 @@ answer is missing or ambiguous, stop and re-ask.
 
 ### 6.2 Verify Watch Status
 
+First, confirm the user's `gh` token has the `notifications` scope —
+without it `/subscription` returns 404 for every repo, whether
+watching or not:
+
+```bash
+gh auth status 2>&1 | grep -i 'notifications'
```
yuezengwu: This check is too broad for the problem you’re trying to solve. `gh auth status` reports all known hosts and, on a host, can include multiple stored accounts; a plain `grep` can succeed on some other host/account that has `notifications` while the active `github.com` account still lacks it. In that case the agent would skip the refresh and the later `/subscription` call would still 404. Please scope the probe to the active `github.com` auth context, e.g. `gh auth status --active --hostname github.com 2>&1 | grep -i 'notifications'`.

yuezengwu (changes requested):
The new scope gate still has a false-positive path: `gh auth status 2>&1 | grep -i 'notifications'` is not scoped to the active `github.com` account. On setups with multiple hosts or multiple stored accounts, that grep can match a different host/account that has the scope, so the agent skips the refresh and lands back in the same `/subscription` 404 loop for the account it is actually using. Please scope the check to the active `github.com` auth context (for example `gh auth status --active --hostname github.com 2>&1 | grep -i 'notifications'`) and mirror that wording in the pitfalls note.

serenakeyitan:
Thanks — good catch. Pushed 045ad06 scoping both the Step 6.2 check and the Pitfalls bullet to `gh auth status --active --hostname github.com` so the grep only sees the active github.com account's token scopes. Verified against my own multi-output `gh auth status` — the scoped form cleanly returns just the active account's block.

yuezengwu (approved):
LGTM — 045ad06 scopes both the Step 6.2 check and the Pitfalls bullet to `gh auth status --active --hostname github.com`, which closes the multi-host / multi-account false-positive path. Thanks for the quick turnaround.

PR #337, opened 2026-04-24
fix(gardener): harden first-tree repo PR review path

bingran-you:
## Summary
- make the first-tree repo's own sync workflow build and run the checked-out CLI instead of npm `latest`
- narrow the Claude classifier subprocess so CI skips project/local slash-command settings and tools it doesn't need
- add classifier coverage for the deterministic CI spawn flags

## Verification
- `pnpm exec vitest run tests/gardener/gardener-claude-cli-classifier.test.ts tests/gardener/gardener-install-workflow.test.ts`
- `pnpm typecheck`
- `pnpm build`
- `node dist/cli.js gardener comment ... --dry-run`

### DIFF
--- a/.github/workflows/first-tree-sync.yml
+++ b/.github/workflows/first-tree-sync.yml
@@ -61,24 +61,23 @@ jobs:
           GIT_ASKPASS="$askpass_script" git -c credential.helper= clone --depth 1 "$tree_repo_url" "$tree_repo_dir"
           rm -f "$askpass_script"
 
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
           node-version: "22"
+          cache: "pnpm"
 
-      - name: Install CLIs (first-tree + claude)
-        # `npx -p first-tree first-tree …` (the gardener-install-workflow
-        # template's default) is flaky on Node 22 / newer npm — it
-        # fails with `first-tree: not found` after install. Install both
-        # globally instead.
-        #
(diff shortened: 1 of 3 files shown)


.github/workflows/first-tree-sync.yml
```diff
@@ -61,24 +61,23 @@ jobs:
...
+        uses: pnpm/action-setup@v4
+
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
           node-version: "22"
+          cache: "pnpm"
+
+      - name: Install repo dependencies
+        run: pnpm install --frozen-lockfile
```
yuezengwu, line 74: Running `pnpm install --frozen-lockfile` here means an unreviewed same-repo PR can execute arbitrary lifecycle/build code before review, with `TREE_REPO_TOKEN` and `GH_TOKEN` already present at the job level and `CLAUDE_CODE_OAUTH_TOKEN`/`ANTHROPIC_API_KEY` available once `node dist/cli.js gardener comment` runs. The previous workflow deliberately executed the published `first-tree` package instead of the PR checkout; switching to `pnpm install` + `pnpm build` + `node dist/cli.js` turns this into a secret-exfiltration path for any contributor who can open an in-repo branch PR. We need to preserve the trusted released binary here, or move this execution to a workflow/event that does not expose secrets to unreviewed PR code.

yuezengwu (changes requested):

Requesting changes for one blocking issue: this workflow now executes the PR checkout itself inside a secret-bearing `pull_request` job, which breaks the previous trust boundary.

Issue #338, opened 2026-04-24
[gardener/classifiers] raise DIFF_CAP and filter noise — 20KB is 6% of Haiku's window


serenakeyitan:
## Problem

Both classifiers (`anthropic.ts:40`, `claude-cli.ts:36`) cap diff input at 20,000 bytes with no content awareness. Two issues compound:

1. **20KB is far too small.** The default model is `claude-haiku-4-5` (200K context window). Current total prompt is ~55KB (~13K tokens) — only 6% of available context. The cap truncates real PRs at 300-400 lines of code. A typical feature PR is 800-1500 lines; a refactor or migration is multiples of that.

2. **Lockfile noise eats the budget.** A PR that regenerates `pnpm-lock.yaml` blows through 20KB of lock diff before a single line of real code reaches the classifier. `tree sync` already solved this (`sync.ts:665` `formatPrDiffForPrompt`) by filtering noise first, then truncating.

### 1. Raise caps

| Constant | Current | New | Share of 200K window |
|---|---|---|---|
| `DIFF_CAP` (both classifiers) | 20,000 | **200,000** | ~25% |
| `DIGEST_BUDGET_BYTES` (`tree-digest.ts:25`) | 30,000 | **100,000** | ~12% |
| Combined with system + metadata | ~55KB | ~310KB | ~38% of Haiku's 200K |

Leaves ~60% of window for prompt-caching overhead, tokenizer variance on mixed Chinese/code content, and Anthropic's soft ~180-190K prompt-length threshold. Works for Sonnet/Opus upgrades without re-tuning.

### 2. Filter noise before truncating

Drop from both classifiers' diff input:

- `*.lock`, `package-lock.json`, `pnpm-lock.yaml`, `yarn.lock`, `Cargo.lock`, `poetry.lock`, `Gemfile.lock`
- `*.min.js`, `*.min.css`, `*.map`
- `dist/`, `build/`, `node_modules/`, `coverage/`, `__pycache__/`, `out/`

Regex list already exists in `sync.ts:626-630`. Extract to shared `engine/classifiers/diff-filter.ts` so sync and both classifiers share one source of truth.

### 3. Single default — no flag

One set of numbers for all scenarios (daemon sweeps, CI, manual invocations). Keeping it simple; we can add an override later only if a real workload demands it.

## Cost implication

[...]

PR #339, opened 2026-04-24
fix(gardener/classifiers): raise DIFF_CAP and filter noise

serenakeyitan:
## Summary

- Raise `DIFF_CAP` from 20KB → 200KB in both `anthropic.ts` and `claude-cli.ts` — 20KB is ~6% of Haiku 4.5's 200K window and truncates typical feature PRs at ~300-400 lines.
- Raise `DIGEST_BUDGET_BYTES` from 30KB → 100KB in `tree-digest.ts`.
- Extract `DIFF_NOISE_PATTERNS` from `sync.ts` into a new shared `engine/classifiers/diff-filter.ts`. Both classifiers now strip lockfiles, `dist/build/out/coverage/node_modules/__pycache__` hunks, and minified/map/snap artifacts BEFORE applying the byte cap, so the cap bounds real code instead of noise.
- Skip the entire `## Diff` section when filtering leaves nothing (e.g. lockfile-only PRs now get a clean prompt instead of an empty code fence).

Total prompt input rises from ~55KB to ~310KB (~38% of Haiku's 200K window), leaving headroom for prompt-caching stability, tokenizer variance on mixed Chinese/code content, and Anthropic's soft ~180-190K prompt-length threshold.

Refs #338.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test` — 1203 passed, 0 failed, 51 skipped (no regressions)
- [x] New `tests/gardener/gardener-diff-filter.test.ts` — 11 cases covering lockfile drop, dist/ drop, minified artifacts, all-noise PRs (empty output), fail-open on malformed hunks, and real-code-only preservation
- [ ] Manual: run `first-tree gardener comment --pr N --repo o/r` on a lockfile-heavy PR locally and confirm the verdict is grounded in real code, not lock diff


yuezengwu (approved):
Reviewed the diff and ran the new tests locally (11/11 passing, typecheck clean). The fix is well-scoped: filtering noise **before** the byte cap is the right ordering, the ~38% of Haiku's 200K window for total prompt input leaves sane headroom for prompt-caching + tokenizer variance, and the new `filterDiffNoise` tests cover the important shapes (lockfiles, dist hunks, all-noise → empty, fail-open on malformed headers, real-code preservation). Guarding the `## Diff` section so an all-noise PR gets no empty code fence is a nice touch.

One thing that does not match the PR description, worth a small follow-up:

- The description says **"Extract `DIFF_NOISE_PATTERNS` from `sync.ts` into a new shared `engine/classifiers/diff-filter.ts`"**, but `src/products/gardener/engine/sync.ts` (lines 678–686) still defines its own private `DIFF_NOISE_PATTERNS` and `isDiffNoise`. The two lists are currently identical, but there are now two sources of truth to keep in sync. `formatPrDiffForPrompt` operates on `PrFileChange[]` (file-level) while the new helper operates on raw unified-diff text, so they can't share the whole function — but `sync.ts` could still `import { DIFF_NOISE_PATTERNS, isDiffNoise } from "./classifiers/diff-filter.js"` and drop its local copy. Not a blocker, but worth closing the loop so the next pattern tweak doesn't have to be made in two files.

Minor/optional observations:

- The `b/(.+)$` header regex will misbehave on filenames containing ` b/` (e.g. a path like `foo b/bar.ts`). Extremely unlikely in real repos; fail-open means the worst case is "noise slips through," so not worth fixing unless you want belt-and-suspenders.
- The split on `/(?=^diff --git )/m` assumes that literal string never appears at the start of a line inside a patch body. Fine for `gh pr diff` output in practice; just noting for anyone who runs into a diff-of-a-diff edge case later.

Approving — the duplication note is the only thing I'd actually want addressed, and a follow-up PR is fine if you prefer to keep this one tight.

PR #351, opened 2026-04-28
feat(gardener): use the actual gardener login in `@<bot> fix` mentions

serenakeyitan:
## Summary

The PR-body footer that gardener-sync emits hard-codes `@gardener fix`. That mention only notifies a GitHub user named "gardener" — but the actual bot identity is whatever account runs gardener (resolved via `gh api user` or `GARDENER_LOGIN`). On any deployment where the bot is a maintainer's personal account, reviewers' `@gardener fix` comments notify nobody. The respond engine had the same issue with its static regex.

## Changes

- **Export `resolveGardenerLogin(shell)`** so sync.ts can reuse it.
- **Add `buildFixCommandRegex(gardenerLogin)`** — builds the dynamic regex with regex-metachar escaping and a fallback to `@gardener` when the login is empty/whitespace (preserves prior behavior for default-named accounts).
- **Wire dynamic regex into respond's fix-detection** (replaces `GARDENER_FIX_RE`).
- **Wire dynamic login into sync's PR-body template** — resolve once at top of `runSync` (env override → `gh api user`), thread through `prepareProposalGroup`, emit `\`@${gardenerLogin || "gardener"} fix\`` in the body.
- **Update skip diagnostic** in respond to print the actual login.

## Test plan

- [x] New `buildFixCommandRegex` unit tests: actual-login match, empty-login fallback, metachar escaping.
- [x] Existing self-loop guard test updated to use `@serenakeyitan fix` (matches its `GARDENER_LOGIN=serenakeyitan` setup) — confirms the literal `@gardener fix` no longer matches a custom login.
- [x] Sync golden snapshots regenerated to include the new `gh api user --jq .login` call at the top of `runSync`.
- [x] Full test suite: 1206 passed (3 new), 0 failed.
- [x] `tsc --noEmit` clean.

## Discovered while

Caught while running gardener locally on `serenakeyitan/paperclip-tree`. The PR-body footer's `@gardener fix` was a dead mention since the bot is `serenakeyitan` — manual local patch worked but kept getting overwritten by `npm i -g first-tree`.

### DIFF
--- a/src/products/breeze/engine/daemon/runner.ts
+++ b/src/products/breeze/engine/daemon/runner.ts
@@ -371,6 +371,14 @@ export function buildAgentEnv(
     BREEZE_BROKER_DIR: request.ghBrokerDir,
     BREEZE_SNAPSHOT_DIR: request.snapshotDir,
     BREEZE_TASK_DIR: request.taskDir,
+    // Surface the daemon's resolved GitHub login so subprocesses (notably
+    // gardener-respond's `@<bot> fix` regex) match the actual bot name
+    // instead of the literal "gardener". Without this, snapshot-mode
+    // runRespond() falls back to "gardener" and reviewers' mentions
+    // silently no-op on any deployment using a custom bot account.
+    // Reuses the existing GARDENER_USER contract from comment.ts /
+    // install-workflow.ts rather than introducing a second env name.
+    GARDENER_USER: request.identity.login,
   };
 }
 
--- a/src/products/gardener/engine/respond.ts
+++ b/src/products/gardener/engine/respond.ts
@@ -268,7 +268,7 @@ export async function resolveGardenerLogin(shell: ShellRun): Promise<string> {
  *
  * The bot's GitHub login isn't always literally "gardener" — it's whatever
  * account is configured to run gardener (resolved via `gh api user` or
- * `GARDENER_LOGIN`). The regex must match the actual login so reviewers'
+ * `GARDENER_USER`). The regex must match the actual login so reviewers'
  * `@<bot> fix` mentions trigger the fix path. Falls back to literal
  * "gardener" when the login is unknown so existing behavior is preserved
  * for default-named accounts.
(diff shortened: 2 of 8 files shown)


serenakeyitan:
@bingran-you could you take a look? Caught locally on `serenakeyitan/paperclip-tree` — the `@gardener fix` mention in every sync PR's body is a dead mention since my bot identity is `serenakeyitan`, not `gardener`. Same issue in the respond engine's static regex. PR makes both dynamic from `gardenerLogin` (resolved from env or `gh api user`), with a fallback to the literal `gardener` for default-named accounts. 1206 tests pass.

bingran-you (changes requested):
Requesting changes because the main bug looks only half-fixed on the actual breeze-runner path. runRespond() still skips gh api user in snapshot mode and only consults env.GARDENER_LOGIN (src/products/gardener/engine/respond.ts:876-881), but breeze-runner buildAgentEnv() does not export GARDENER_LOGIN at all (src/products/breeze/engine/daemon/runner.ts:364-374). In the real review-request flow that means gardenerLogin == "", buildFixCommandRegex() falls back to @gardener, and @<actual-login> fix comments from the sync PR body still will not trigger respond unless someone manually injects that env var. The new tests only cover snapshot mode with GARDENER_LOGIN already present, so they miss the deployed path. Please wire the login through the runner environment, or reuse the existing GARDENER_USER plumbing instead of introducing a second env contract, and add a snapshot-mode test that exercises a custom bot login without manual gh api user lookup. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

serenakeyitan:
Pushed fix addressing the runner-path bug:

- **Root cause confirmed.** `runRespond()` only consulted `env.GARDENER_LOGIN`, but `buildAgentEnv()` in `daemon/runner.ts:364-374` never exported it — so on the real breeze-runner path `gardenerLogin == ""` and `buildFixCommandRegex()` fell back to the literal `gardener`. The previous tests passed only because they pre-set the env manually.

- **Fix per your suggestion (reuse existing contract, not a second one).** `buildAgentEnv()` now exports `GARDENER_USER` from `request.identity.login` — the daemon's already-resolved login from `DaemonIdentity` (which is what `gh auth status` returns). `respond.ts` and `sync.ts` now read `GARDENER_USER` as the canonical name (matching `comment.ts:1719` and `install-workflow.ts:242`). `GARDENER_LOGIN` is honored as a back-compat alias so older deployments keep working.

- **New tests covering the deployed path** (your specific ask):
  - `tests/gardener/gardener-respond.test.ts` — snapshot mode with `GARDENER_USER=custom-bot-acct`, asserts no `gh api user` call fires AND the `@custom-bot-acct fix` mention is recognized (hard-fails on `gh api user`).
  - Same file — back-compat test confirming `GARDENER_LOGIN` still works when `GARDENER_USER` is unset.
  - `tests/breeze/breeze-daemon-runner-agent.test.ts` — asserts `buildAgentEnv()` exports `GARDENER_USER` from `request.identity.login`.

All 1209 tests pass; typecheck clean.

Ready for re-review. cc @bingran-you

serenakeyitan (commented):
Please proceed with this approach — confirmed via #366 E2E test.

serenakeyitan:
Note: the previous PR review at https://github.com/agent-team-foundation/first-tree/pull/351#pullrequestreview-4214149312 was generated by an interrupted automated end-to-end test for #366, NOT a real review. GitHub's API does not allow dismissing `COMMENTED` reviews (only `APPROVED`/`CHANGES_REQUESTED` are dismissible). Apologies for the noise. cc @bingran-you

Issue #321, opened 2026-04-23
[breeze] Pass --auto when opening PRs so approved PRs merge themselves on opted-in tree repos


serenakeyitan:
## Problem

Breeze opens PRs against tree repos, bingran (or another reviewer) approves, breeze sets \`breeze:done\`, and then... the PR sits there forever. The loop dead-ends.

Concrete example: \`serenakeyitan/paperclip-tree\` currently has 20+ PRs that are \`reviewDecision: APPROVED\`, \`mergeable: MERGEABLE\`, \`mergeStateStatus: CLEAN\`, labeled \`breeze:done\`. Nothing merges them.

The README onboarding promises "source PR → tree issue → breeze pickup → draft-node PR" as an end-to-end loop, but the loop isn't closed.

## Proposed fix

When breeze opens a PR against a tree repo, also call \`gh pr merge --auto --squash --delete-branch\` on it. GitHub's native auto-merge takes over from there:

- If the target repo has \"Allow auto-merge\" **enabled** in Settings → GitHub merges the PR the moment review + required checks are green.
- If the target repo has it **disabled** (the default) → the \`--auto\` call silently no-ops, PR sits as today. Zero behavior change for any repo that hasn't opted in.
- If a reviewer later requests changes → GitHub cancels the pending auto-merge. Breeze addresses, re-requests review, and re-calls \`--auto\` on the next sweep.

## Why this shape

- **One switch, one place.** The tree-repo owner turns on \"Allow auto-merge\" in their repo Settings. No YAML config, no parser changes, no PR into first-tree to enable it.
- **Per-tree by construction.** \`paperclipai/paperclip-tree\` can enable it; \`agent-team-foundation/first-tree\` leaves it off; each OSS user decides for themselves. The switch lives where the authority lives.
- **Review is still the gate.** \`--auto\` respects GitHub's required-reviews setting. Breeze never merges without approval.
- **Nothing new to maintain.** No \`auto_merge\` config key, no \`allowed_authors\` list, no \`max_files_changed\` check. Just lean on the GitHub primitive.

## Alternatives considered

- **Breeze daemon sweep merges approved PRs itself** (with per-tree YAML config). Gives more knobs (\`allowed_authors\`, size gates) but requires new config plumbing and a PR-to-first-tree to enable. Too heavy for an OSS tool whose users just want their loop to close.
[...]

PR #352, opened 2026-04-28
fix(sync): make `gh pr merge --auto` opt-in (default off)

serenakeyitan:
## Summary

Sync used to call `gh pr merge --auto --squash --delete-branch` on every freshly-opened tree PR. The comment on the call assumes branch protection will gate it — but **on a repo without protection, `--auto` merges immediately**. There's no waiting, no review.

19 sync PRs on `serenakeyitan/paperclip-tree` were merged this way today — zero approvals across all of them. The reviewer (bingran-you, via breeze) had no chance to review; sync opened and merged in the same instant. Confirmed by checking the merge timestamps and `merged_by` (matches the sync run, no APPROVED reviews on file).

## Fix

Auto-merge is now **opt-in via config**, defaults off:

```yaml
modules:
  sync:
    enabled: true
    auto_merge: true   # only safe if your tree repo requires approvals
```

- New `OpenTreePrOpts.autoMerge?: boolean` (default false). Skips the `gh pr merge --auto` step entirely when false.
- New `ModuleToggle.auto_merge?: boolean`. Sync reads `modules.sync.auto_merge` and threads it through `pushAndCreatePr(s)` → `openTreePr`.
- Doc comments call out the unsafe-without-protection caveat clearly.

## Behavior change for existing users

If your deployment was relying on the default auto-merge, you now need to opt in. **Don't** opt in unless your tree repo has branch protection that requires approvals — the config flag only makes the choice explicit, it doesn't make `--auto` safe on an unprotected repo.

## Test plan

- [x] New: "does NOT queue auto-merge by default" — verifies the no-merge call sequence on fresh deployments.
- [x] Existing auto-merge tests updated to opt in with `autoMerge: true`. All 4 still pass.
- [x] Sync golden snapshot for `two-pr-apply` regenerated (no more `gh pr merge` in default call sequence).
- [x] `pnpm test` → 1204 passed, 0 failed.
- [x] `pnpm typecheck` clean.

## Discovered while

Running gardener locally on `serenakeyitan/paperclip-tree`. Audit of recently-merged sync PRs showed 19/20 had `[]` for non-gardener APPROVED reviews, all merged within seconds of being opened by the same sync run.

### DIFF
--- a/src/products/gardener/engine/runtime/config.ts
+++ b/src/products/gardener/engine/runtime/config.ts
@@ -28,6 +28,24 @@ import { join } from "node:path";
 
 export interface ModuleToggle {
   enabled: boolean;
+  /**
+   * Sync-only: queue GitHub native auto-merge on freshly-opened tree
+   * PRs. **Defaults to false.** Setting `true` is safe ONLY when the
+   * tree repo has branch protection that requires approvals/checks —
+   * `gh pr merge --auto` waits for those before merging. On a repo
+   * without branch protection, `--auto` merges immediately and bypasses
+   * review entirely.
+   *
+   * The flag is read from the same `modules.sync` block:
+   *
+   *   modules:
+   *     sync:
+   *       enabled: true
+   *       auto_merge: true   # only if your tree repo requires approvals
+   *
+   * Ignored on non-sync modules.
+   */
+  auto_merge?: boolean;
 }
 
 export interface GardenerConfig {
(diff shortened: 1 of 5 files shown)


serenakeyitan:
@bingran-you please take a look — this is the unsafe-default I caught on `paperclip-tree` (19 sync PRs merged with zero approvals). Tests pass (1204/1204). The change is API-compatible at the OpenTreePrOpts level (new optional field) but **does change default sync behavior** — opt-in required for repos that previously relied on auto-merge.

bingran-you (approved):
LGTM — approving.

This is the right fix. The bug is genuinely scary: `gh pr merge --auto` on an unprotected tree repo merges immediately, and the prior comment on the call site implicitly assumed branch protection would always be in place. Defaulting `autoMerge` to false moves the unsafe path behind explicit opt-in, and the JSDoc on both `OpenTreePrOpts.autoMerge` and `ModuleToggle.auto_merge` calls out the "still unsafe without branch protection" caveat plainly so opt-in users understand the flag isn't a safety guarantee on its own.

Spot checks I did before approving:
- Threading: `runSync` reads `modules.sync.auto_merge` and passes it through `pushAndCreatePr(s)` → `openTreePr`. The other `openTreePr` caller, `draft-node.ts`, omits the option, which is correct — draft-node PRs are explicitly meant for human review.
- Config parsing: `coerceModuleToggle` keeps the old "missing `enabled` returns undefined" semantics and only adopts `auto_merge` when it's a real boolean (or string `"true"`/`"false"`), matching the rest of the file's coercion style.
- Config location: `loadGardenerConfig(treeRoot)` is consistent with the existing `comment.ts` / `respond.ts` callers (tree-repo `.claude/gardener-config.yaml`).
- Tests: the new "does NOT queue auto-merge by default" case locks in the new contract; the four merge-fallback tests opted in with `autoMerge: true` rather than getting deleted, which preserves coverage of the failure modes (#321). The `two-pr-apply` golden snapshot dropped both `gh pr merge` calls as expected.

Two non-blocking notes for whoever lands this:

1. **Behavior change for existing users.** Anyone relying on the old default loses auto-merge silently after upgrade. Worth a bold callout in the next release notes / CHANGELOG so deployments with proper branch protection know to add `auto_merge: true`. The PR description already makes the contract clear; just want it visible to non-PR-readers too.

2. **The failing `tree-sync` check on this PR is unrelated** — `claude -p exited 1` from the CI gardener-comment step, not a code regression. CI (typecheck + tests) is green.

Nice catch on `paperclip-tree`. This is the kind of "default looks fine in the happy path, eats your reviews in the sad path" issue that's easy to miss until it bites.

Issue #358, opened 2026-05-01
spec(github-scan): auto-revert `github-scan:human` → `new` on new human comment


serenakeyitan:
## Problem

The classifier treats `github-scan:human` as a one-way trapdoor: once an item is labeled `human`, the daemon never re-evaluates, even if the human comments back with a decision.

This breaks the natural handoff flow:

1. Agent escalates: *"I need a human to decide X"* → labels `github-scan:human`.
2. Human reads, replies in the issue with the decision (e.g. *"go ahead with option A"*).
3. **Human expects the agent to read the decision and continue working.**

Step 3 silently fails today. The only way out of `human` is for the human to *also* manually strip the `github-scan:human` label — two actions for one logical handoff. People forget, so agents silently sit idle on resolved escalations.

### Concrete example

[first-tree-website#10](https://github.com/agent-team-foundation/first-tree-website/issues/10) — the human commented `@serenakeyitan make it done`. The daemon polled, saw the new activity, but classified the item `human` and skipped it because the label was still set.

## Proposed fix (Option A: auto-revert on new human comment, with guard)

When the daemon polls an item classified `human` and observes a new human comment posted **after** the label was applied:

1. **Strip** `github-scan:human` from the item.
2. Item re-classifies as `new` (default for OPEN items with no `github-scan:*` label).
3. Daemon picks it up on the next dispatch cycle.

### Guard rails (avoid spurious re-fires)

- Comment author MUST NOT be the agent itself (the daemon's identity).
- Comment body length > 20 chars (filter out `👍`, `ok`, `thx`-only acks).
- Reactions alone do NOT count as a comment.
- The comment's `created_at` MUST be strictly after the label's `created_at` (or the label-event timestamp from the issue timeline).

### Out of scope for this issue

- A separate `/handoff` magic phrase or bot-mention trigger (Option B). The auto-revert covers ~95% of the natural flow; explicit triggers can be added later if Option A produces false positives in practice.
- Tray-driven handoff (clicking a row in the menu-bar tray to also strip the label). Tracked separately once `apps/tray-mac` lands.

## Why this matters now

The friction is invisible while only a handful of human-labeled items exist and the same person who set the label is also acting on them. As soon as the menu-bar tray ships and the queue becomes glanceable, more `human` items will accumulate and the *"reply in GitHub vs. also remember to strip the label"* friction will compound — exactly the wrong UX as we're promoting `github scan`.

## Acceptance criteria

[...]

PR #309, opened 2026-04-23
fix(breeze): support gh auth status without json

bingran-you:
Fixes #308.

## Summary
- replace daemon identity resolution's `gh auth status --json hosts` dependency with plain `gh auth status --active --hostname <host>` parsing
- normalize quoted and unquoted scope output from `gh auth status`
- add a regression test that asserts the resolver no longer passes `--json`

## Testing
- `pnpm exec vitest run tests/breeze/breeze-daemon-identity.test.ts`
- `pnpm typecheck`
- `pnpm build`
- isolated smoke: `BREEZE_DIR=$(mktemp -d)` then `node dist/cli.js breeze doctor --allow-repo agent-team-foundation/first-tree`

PR #359, opened 2026-05-02
fix(github-scan): auto-revert github-scan:human on new human comment (#358)

serenakeyitan:
Refs #358.

## Summary

When the daemon polls an item classified `human` and a qualifying human comment is observed *after* the `github-scan:human` label was applied, the daemon strips the label so the (unchanged) classifier naturally re-derives `new` and the dispatcher picks it up on the next cycle.

This closes the silent-failure handoff documented in #358 (e.g. agent-team-foundation/first-tree-website#10) where a human replies in the issue and reasonably expects the agent to read the decision and continue working — instead of having to also manually strip the label.

## Implementation

- New module `packages/github-scan/src/github-scan/engine/runtime/auto-revert.ts` exporting:
  - `shouldAutoRevertHuman(input)` — pure decision function applying the four guards.
  - `autoRevertHumanLabels(entries, deps)` — driver that fetches the timeline + comments, calls the existing `gh.removeLabel`, and mutates `entry.labels` in place.
- Wired into `pollOnce` (`packages/github-scan/src/github-scan/engine/daemon/poller.ts`) **between** `enrichWithLabels` and `classifyEntries` — so when guards pass, classification on the same cycle re-derives `new` (no extra round trip needed).
- `runner-skeleton.ts` passes the daemon's resolved `identity.login` as `agentLogin`. When identity resolution fails (degraded mode), auto-revert is skipped — fail-safe.
- The classifier itself is intentionally untouched. Auto-revert is purely an action layer.
- Spurious-fire prevention: we mutate `entry.labels` in the same cycle, and on the *next* cycle the live GraphQL label-set no longer contains `github-scan:human`, so the entry skips the auto-revert branch entirely. No per-poll tombstone needed.
- Updated agent prompt in `daemon/runner.ts` to document the `human → new` transition so agents know they can stop on `github-scan:human` and trust the daemon to re-queue on reply.

## Guards (from issue #358)

| # | Guard | How tested |
|---|---|---|
| 1 | Author MUST NOT be the agent | (a) `own-comment ignored` test |
| 2 | Body length > 20 chars | (b) `short ack ignored` test (`"ok thx"`, `"👍"`) |
| 3 | Reactions alone do NOT count | (c) `reaction ignored` test (empty body shape) |
| 4 | `created_at` strictly after label timestamp | (e) `pre-label comment` test + a separate `==` strict-inequality test |
| — | All guards pass → revert fires | (d) `genuine reply triggers revert` test |

## Acceptance-criteria coverage

`packages/github-scan/tests/github-scan/github-scan-auto-revert.test.ts` — 11 tests, 5 acceptance + 2 extra guard edges + 4 driver tests:

- [x] (a) own-comment ignored
- [x] (b) short ack ignored
- [x] (c) reaction ignored
- [x] (d) genuine reply triggers revert
- [x] (e) pre-label comment does NOT trigger revert
- [x] driver: end-to-end `gh.removeLabel` is called and `entry.labels` is mutated when guards pass
- [x] driver: skips items not labeled `github-scan:human`
- [x] driver: degrades safely when label-event timestamp can't be fetched

## Validation

- [x] `pnpm -r test` — 522 tests pass across 38 files
- [x] `pnpm lint`
- [x] `pnpm typecheck`

cc @bingran-you for review.


bingran-you (approved):
LGTM overall — clean separation between the pure decision function and the I/O-bound driver, all five acceptance cases from #358 are covered, and the in-place `entry.labels` mutation neatly avoids the need for a per-cycle tombstone. A few non-blocking observations:

**Pagination cap (worth a follow-up, not a blocker)**

`fetchHumanLabelAppliedAt` and `fetchIssueComments` both fetch only the first page (`per_page=100`) of the timeline / comments respectively (`packages/github-scan/src/github-scan/engine/runtime/auto-revert.ts:99,145`). For a long-lived issue with > 100 timeline events between the `labeled` event and the head of the timeline, the latest `github-scan:human` `labeled` event could fall off the page and `fetchHumanLabelAppliedAt` would return `null` — auto-revert would then be silently skipped (a warning fires, but the human handoff stays stuck). Same shape for the comments fetch. Realistic for typical issues; could bite for noisy long-running threads. Could either page or — for the timeline specifically — query in reverse via `?page=N&per_page=100` only when the first page doesn't contain the event.

**Silent `removeLabel` failure**

`gh.removeLabel` swallows non-zero exits (`runtime/gh.ts:174`). If the REST call fails after we've mutated `entry.labels` in place, the same cycle will (correctly) re-classify as `new` and dispatch — but on the next cycle the GraphQL re-read will still show `github-scan:human`, the auto-revert path will re-fire, and the agent gets re-dispatched. Probably tolerable (the agent will idempotently respond to the same comment), but worth thinking about whether the in-memory mutation should be conditional on `removeLabel` actually succeeding. A quick check on `removeLabel`'s return + early-exit on failure would close this.

**Body length threshold**

`comment.body.trim().length <= AUTO_REVERT_MIN_BODY_CHARS` (≤ 20) means *strictly more than* 20 characters qualifies, which matches the PR description and the issue spec. Just noting in case anyone reads the constant name and assumes "at least 20".

**Nit — agent-prompt wording**

The new prompt line in `runner.ts:200` says "anyone other than {login}" but the actual guard is case-insensitive on the login *and* requires > 20 chars *and* requires post-label timestamp. If an agent ever reasons about whether to set `github-scan:human` based on this prompt, the simplification is fine; if they ever debug why a revert *didn't* fire, the abbreviated description could mislead. Minor.

Approving — none of the above blocks merge. The pagination one is the only one I'd consider opening a follow-up issue for once you've seen this in production for a bit.

serenakeyitan:
## End-to-end verification

Ran a real-GitHub smoke test of this PR against a separate daemon on `:7879` (production daemon untouched on `:7878`), fresh `GITHUB_SCAN_DIR=/tmp/scan-test-pr359`, `--dry-run` to avoid spawning competing agents.

**Test target:** [`agent-team-foundation/first-tree#309`](https://github.com/agent-team-foundation/first-tree/pull/309) — PR present in the daemon's notifications inbox, no `github-scan:human` at start.

**Test setup note:** my only `gh` identity is `serenakeyitan`, so a real human-vs-agent split was simulated by setting `agentLogin` to a non-`serenakeyitan` value at the `runPoller` call site (one-line patch reverted after the test). Everything downstream of `agentLogin` — `fetchHumanLabelAppliedAt`, `fetchIssueComments`, the four guards in `shouldAutoRevertHuman`, and the `gh.removeLabel` call — ran live against real GitHub.

### Positive case (auto-revert fires)
- Applied `github-scan:human` at `2026-05-02T01:03:28Z` ([timeline](https://github.com/agent-team-foundation/first-tree/pull/309))
- Posted qualifying comment (70+ chars) at `2026-05-02T01:03:32Z`: [`#issuecomment-4362475865`](https://github.com/agent-team-foundation/first-tree/pull/309#issuecomment-4362475865)
- Daemon poll observed:
  ```
  WARN: auto-reverted github-scan:human on 2 item(s) (issue #358)
  github-scan: polled 128 notifications (7 new)
  ```
- Post-poll labels on #309: `breeze:human` only — `github-scan:human` stripped.
- ✅ PASS

### Negative case (guards hold)
- Re-applied `github-scan:human` at `2026-05-02T01:07:38Z`
- Posted short comment "ok" (2 chars) at `2026-05-02T01:07:42Z`: [`#issuecomment-4362485804`](https://github.com/agent-team-foundation/first-tree/pull/309#issuecomment-4362485804)
- Next poll cycle: no auto-revert warning, no label change.
- Post-poll labels on #309: `breeze:human, github-scan:human` (label preserved as expected — guard 2 rejected the short body).
- ✅ PASS

### Real-world surprise worth noting
The poller candidate set comes only from `/notifications`. Items the operator isn't subscribed to (e.g. `first-tree-context#197`, which I tried first) never enter the inbox and so never reach auto-revert, even when they carry `github-scan:human`. Not a regression introduced by this PR — but worth knowing: auto-revert reaches exactly what the daemon polls. Pure unit logic in `auto-revert.ts` is correct; downstream visibility is bounded by the notifications feed.

Test pollution cleaned: `github-scan:human` removed from #309, daemon stopped, `/tmp/scan-test-pr359*` deleted, source patch reverted (`git diff` clean).

cc @bingran-you

Issue #364, opened 2026-05-02
fix(github-scan): check `removeLabel` success before mutating local inbox state


serenakeyitan:
## Problem

`auto-revert.ts:249-251` mutates `entry.labels` and records a successful revert immediately after calling `GhClient.removeLabel`. But `removeLabel` in `gh.ts:169-184` intentionally swallows every `gh` failure — a transient API error, 403, or rate-limit returns silently as if it succeeded.

Result: the local inbox reclassifies to `new` and the dispatcher picks the item up, but **GitHub still shows `github-scan:human`**. Tray and dashboard disagree with the actual GitHub state. Confusing symptoms ("I labeled it human but the agent kept working on it").

## Proposal

Make `removeLabel` return a success/failure signal (or throw on non-zero exit), then have `auto-revert.ts` early-exit on failure without mutating `entry.labels` or recording a revert. Next poll cycle will retry naturally.

## Acceptance criteria

- [ ] `gh.removeLabel` returns a result indicating success/failure (boolean or `Result<void, Error>` — match existing patterns in `gh.ts`).
- [ ] `auto-revert.ts` does NOT mutate `entry.labels` or record the revert when `removeLabel` fails.
- [ ] Unit test: `removeLabel` failure → `entry.labels` unchanged, no revert recorded, daemon will retry next cycle.
- [ ] Existing 11 auto-revert tests still pass.

## Refs

- Spec: #358
- Origin: #359 review feedback ([reviewer comment](https://github.com/agent-team-foundation/first-tree/pull/359#pullrequestreview-...)), re-raised as blocker on #361.

cc @bingran-you

serenakeyitan:
Implemented and shipped via #368.

PR #368, opened 2026-05-02
fix(github-scan): early-exit auto-revert when removeLabel fails (#364)

serenakeyitan:
## Summary

`gh.removeLabel` previously swallowed every gh failure as silent success, and `auto-revert.ts:249-251` mutated `entry.labels` and recorded a successful revert immediately after the call. A transient API error, 403, or rate-limit would leave the local inbox reclassifying to `new` while GitHub still showed `github-scan:human` — tray and dashboard disagreed with actual GitHub state.

## Implementation

- **`packages/github-scan/src/github-scan/engine/runtime/gh.ts`** — `removeLabel` now returns `boolean` (`true` on gh exit 0, `false` otherwise). Boolean was chosen over throwing to keep `status-manager.ts`'s "swallow + continue" caller working without try/catch noise; the harmless "label not on item" failure mode is still represented as `false`.
- **`packages/github-scan/src/github-scan/engine/runtime/auto-revert.ts`** — checks the result. On `false`: log a warning, `continue` without mutating `entry.labels` and without recording the revert. Next poll cycle's GraphQL re-read will retry naturally.
- **`packages/github-scan/tests/github-scan/github-scan-auto-revert.test.ts`** — new failure-path test asserts `entry.labels` is unchanged, `out.reverted` is empty, and a warning is recorded when `removeLabel` returns `false`. Existing happy-path test updated to mock `true`.

## Verification

- `pnpm -r test` — 448 + 77 tests pass (including all 11 existing auto-revert tests + 1 new)
- `pnpm lint` — 0 warnings, 0 errors
- `pnpm typecheck` — clean
- **No live E2E run** — per user direction, forcing a transient GitHub API failure mid-cycle isn't reliable; unit-test-with-mocked-failure is the right level here.

## Test plan
- [x] Unit tests cover the new failure path
- [x] All existing auto-revert tests still pass
- [x] `pnpm -r test`, `pnpm lint`, `pnpm typecheck` all green

Refs #364

cc @bingran-you


bingran-you (approved):
## Review — LGTM

Focused, well-scoped fix that delivers exactly what issue #364 asked for. Approving.

### What works well
- **API choice is right.** `removeLabel: void → boolean` is the minimal-surface change. Throwing would have forced `status-manager.ts:229`'s deliberate swallow pattern into try/catch noise, and a `Result<>` type would be over-engineered for a binary outcome. The docstring at `gh.ts:142-150` explains this trade-off clearly for the next reader.
- **Caller correctness verified.** Only two callers exist — `auto-revert.ts:254` (now checks) and `status-manager.ts:229` (intentionally swallows; behavior preserved since `void` callers can drop a `boolean` return). No silent breakage.
- **State-machine reasoning is sound.** On `false`, the early `continue` skips both the `entry.labels` mutation and the `reverted.push(entry.id)`, so the next poll cycle's GraphQL re-read sees `github-scan:human` still present and naturally retries. Matches the spec note at `auto-revert.ts:207-212`.
- **Test covers the right invariants.** The new test at `github-scan-auto-revert.test.ts:253-280` asserts all four properties: `removeLabel` was called, `entry.labels` retains the label, `out.reverted` is empty, and a warning is recorded. Existing happy-path test correctly updated to `mockReturnValue(true)`.
- **Warning message is actionable** — `"removeLabel failed for ${repo}#${number}; will retry next cycle"` tells operators both what failed and what the system will do about it.

### Minor (non-blocking)
- The unrelated formatting changes in `gh.ts` (one-line constructor signature, collapsed array args in `addLabelWithFallback`/`createLabel`) add diff noise unrelated to the bug fix. Ideally a separate commit, but not worth blocking on for a 3-file PR.
- `github-scan-gh.test.ts:141-145` (`"swallows failure"`) still asserts only that `removeLabel` doesn't throw — it could now also assert the return value is `false` to lock in the new contract. Optional follow-up; the auto-revert test already exercises the false path through the integration boundary.

### Verification
Trusting the PR's stated `pnpm -r test` / `pnpm lint` / `pnpm typecheck` results — couldn't re-run locally (workspace `node_modules` not installed in this review env). The diff is small enough that the unit tests are the right level of coverage here; agreed that forcing a transient `gh` API failure for live E2E isn't reliable.

serenakeyitan:
Thanks for the thorough review @bingran-you — appreciate the API trade-off walk-through, the caller-correctness check on status-manager.ts:229, and the state-machine reasoning on the early continue. Noted on the follow-ups: the unrelated gh.ts formatting churn should have been a separate commit, and tightening github-scan-gh.test.ts:141-145 to assert removeLabel returns false (locking the new contract at the unit boundary) is a good follow-up — will pick that up in a small follow-up PR rather than amending this one. Will merge once main is rebased/refreshed. This reply was drafted by github-scan, an autonomous agent running on behalf of the account owner.

Issue #366, opened 2026-05-02
fix(github-scan): include PR reviews + review comments in auto-revert qualifying-comment scan


serenakeyitan:
## Problem

`auto-revert.ts:92-126` only inspects `/repos/{owner}/{repo}/issues/{n}/comments`. PR review bodies and PR review comments never go through that endpoint — they live at:

- `/repos/{owner}/{repo}/pulls/{n}/reviews` (review bodies: \"Approve\" / \"Request changes\" / \"Comment\" submissions)
- `/repos/{owner}/{repo}/pulls/{n}/comments` (inline line-by-line review comments)

Result: a reviewer can reply via the PR Review UI (the most natural reply path for code review) and the thread stays stuck in `github-scan:human`. Auto-revert effectively doesn't work for PR review replies — which is the **dominant reply surface** for PRs.

The rest of the daemon already treats reviews as visible activity (see `gh-client.ts:226-267`), so the auto-revert path should cover the same surfaces.

## Proposal

Extend the qualifying-comment scan to also fetch:
- PR review bodies (where `body` is non-null)
- PR inline review comments

For each, apply the same guards as today's flow:
- author ≠ resolved agent identity
- body length > 20 chars (trimmed)
- non-empty (reactions excluded — N/A for these endpoints)
- `created_at` strictly > label-event timestamp

For PRs, fetch all three sources (issue comments + reviews + review comments), pick the latest qualifying one, and run the auto-revert if any pass.

Also: **paginate** these new fetches (don't repeat the issue raised in #X — the pagination follow-up issue).

## Acceptance criteria

- [ ] PR review body with > 20 chars by a non-agent user, after the label, triggers auto-revert.
- [ ] PR inline review comment with > 20 chars by a non-agent user, after the label, triggers auto-revert.
- [ ] Issue comments (current behaviour) still trigger auto-revert.
- [ ] Reviews with empty body (just an APPROVE click, no text) do NOT trigger auto-revert.
- [ ] Pagination handled (or explicit follow-up dependency on the pagination issue).
- [ ] Unit tests for each path: issue comment, PR review body, PR review comment, mixed (issue with both kinds of activity).
- [ ] Live E2E: post a PR review reply on a real PR with `github-scan:human`, watch auto-revert fire.

## Refs

- Spec: #358
- Origin: #359 review feedback (raised as blocker on #361 review) — most user-impactful of the three follow-ups.

cc @bingran-you

PR #361, opened 2026-05-02
feat(github-scan): allow declaring agent identity separately from operator (#360)

serenakeyitan:
Refs #360. Surfaced from the live #359 smoke ([test report](https://github.com/agent-team-foundation/first-tree/pull/359#issuecomment-4362488313)) where the operator's `gh auth` user (`serenakeyitan`) was the same as the daemon's agent identity, so every operator comment was filtered as own-comment and auto-revert never fired.

## Summary

Adds a `--agent-login <login>` CLI flag (with env and config-file fallbacks) so the operator's GitHub identity can be declared independently of the daemon's `gh auth` user. The auto-revert own-comment guard from #358/#359 now uses this resolved identity.

## Resolution order

| # | Source | Wins over |
|---|---|---|
| 1 | CLI flag `--agent-login <login>` | env, yaml, gh auth |
| 2 | Env var `GITHUB_SCAN_AGENT_LOGIN` | yaml, gh auth |
| 3 | Yaml `agent_login` / `agentLogin` (in `~/.first-tree/github-scan/config.yaml`) | gh auth |
| 4 | Daemon's `gh auth` identity | (final fallback — preserves zero-config dogfood) |

## Implementation

- `runtime/config.ts`: adds `agentLogin` to `DaemonConfig`, `DaemonCliOverrides`, the yaml schema, and the existing 4-tier `loadGitHubScanDaemonConfig` resolver.
- `daemon/runner-skeleton.ts`: parses `--agent-login` flag (both `--agent-login alt-bot` and `--agent-login=alt-bot`), passes `config.agentLogin ?? identity?.login` into `runPoller` / `runPollerOnce`. Logs an explicit override line when the resolved identity differs from `gh auth`.
- `cli.ts`: surfaces `--agent-login <login>` in `run`, `daemon`, and `start` help text.
- `auto-revert.ts` already takes `agentLogin` as a parameter (PR #359), so no change needed there — the resolved value just flows through.

## Tests

- 6 new config tests (all 4 resolution tiers + camelCase yaml + empty-CLI-no-clobber)
- 2 new `parseDaemonArgs` tests (both `--agent-login` forms + empty-value rejection)
- 1 new auto-revert test verifying guard 1 (own-comment) uses the *resolved* identity, NOT the `gh auth` user — this is the exact #360 scenario
- All existing 11 auto-revert tests + existing config tests still pass
- `pnpm -r test` (531 tests), `pnpm lint`, `pnpm typecheck` all pass

[...]

Issue #365, opened 2026-05-02
fix(github-scan): paginate auto-revert timeline + comments fetches


serenakeyitan:
## Problem

`fetchHumanLabelAppliedAt` (`auto-revert.ts:97-100`) and `fetchIssueComments` (`auto-revert.ts:143-149`) fetch only the first page (`per_page=100`) of timeline events / comments respectively. On long-lived issues with > 100 timeline events, the most recent `github-scan:human` `labeled` event can fall off page 1 and `fetchHumanLabelAppliedAt` returns `null` — auto-revert silently skips and the human handoff stays stuck. Same shape for the comments fetch (101st comment is invisible).

Realistic for noisy long-running threads: a contentious PR with multiple reviewers and several rounds of changes can hit 100 timeline events in weeks.

## Proposal

Page through results until either (a) the relevant event is found, or (b) we've gone back far enough that an older event wouldn't matter. Two reasonable shapes:

**(a) Forward pagination with early exit** — page from page 1 until the label event is found OR `created_at` < label timestamp.

**(b) Reverse pagination (timeline-only)** — start from the last page (`Link: rel="last"` header) and walk backwards. Faster for old labels on long threads but adds a Link-header parse.

For comments, forward pagination with an early-exit on `created_at <= labelAppliedAt` is enough.

## Acceptance criteria

- [ ] `fetchHumanLabelAppliedAt` returns the correct timestamp even when the label-application event is on page 2+.
- [ ] `fetchIssueComments` returns all qualifying post-label comments even when they're on page 2+.
- [ ] Pagination terminates early (don't fetch all 50 pages of a 5,000-comment thread when page 3 is already pre-label).
- [ ] Unit tests with mocked GitHub responses covering: 0-page (empty), 1-page (existing behaviour), 2-page (label-event on page 2), 3+-page (comment on page 3 newer than label on page 2), early-exit behaviour.

## Refs

- Spec: #358
- Origin: #359 review feedback (flagged as non-blocker there, re-raised as blocker on #361 review).

cc @bingran-you

serenakeyitan:
Implemented and shipped via #370.

PR #369, opened 2026-05-02
fix(github-scan): include PR reviews in auto-revert qualifying-comment scan (#366)

serenakeyitan:
## Summary

Auto-revert previously only inspected `/repos/{r}/issues/{n}/comments`, leaving PR review bodies (`/pulls/{n}/reviews`) and PR inline review comments (`/pulls/{n}/comments`) silently invisible. Since PR review replies are the dominant code-review reply path, auto-revert effectively did not fire for the most common case.

For PR entries (`entry.type === "PullRequest"`), the driver now fetches from all three surfaces, concatenates the comments, and feeds them to the same `shouldAutoRevertHuman` decision function. The four guards from #358 are unchanged: author ≠ agent, body length > 20 chars (trimmed), non-empty body, `created_at` strictly > label-event timestamp. Issues retain prior behaviour — the new fetchers are only invoked when `entry.type === "PullRequest"`.

`fetchPrReviews` reads `submitted_at` (the field GitHub returns for review submissions) instead of `created_at`; both timestamps go through the same `Date.parse` guard.

A failed sub-fetch (reviews or review-comments) now produces a soft warning and the driver proceeds with whatever comments were retrieved, so a working issue-comment fetch can still fire on its own — auto-revert never silently strips on a failed fetch.

## Test cases

| # | Case | Expected | Status |
|---|---|---|---|
| 1 | PR review body > 20 chars by non-agent, after label | revert | pass |
| 2 | PR inline review comment > 20 chars by non-agent, after label | revert | pass |
| 3 | PR review with empty body (APPROVE click only) | NO revert | pass |
| 4 | PR with both an issue comment and a review reply | revert | pass |
| 5 | Issue (non-PR) entry — PR fetchers NOT invoked | revert via issue comment only | pass |
| 6 | PR with all three sources, latest qualifying wins | revert | pass |
| 7 | PR review fetch failure does not block issue-comment qualifier | revert + 2 warnings | pass |
| 8 | PR review by the agent itself | NO revert | pass |
| All existing #358 unit tests | unchanged | pass |

`pnpm -r test`: 532 tests pass (455 in `@first-tree/github-scan`, 77 in `apps/cli`).
`pnpm lint`: 0 warnings, 0 errors.
`pnpm typecheck`: clean.

## LIVE E2E

**Blocked.** The E2E plan in the issue requires running the daemon under a non-agent identity so that a PR review posted by the operator (`serenakeyitan`) is treated as a non-agent reply. The current branch resolves the daemon's `agentLogin` strictly from `resolveDaemonIdentity()` (i.e. `gh auth` of the running user) — there is no `--agent-login` flag and no `GITHUB_SCAN_AGENT_LOGIN_OVERRIDE` env var on this branch. PR #361 (which adds the disambiguation flag) is still OPEN, not merged.

Without that plumbing, any PR review I post via `gh` would be authored by the same login the daemon resolves, and the author guard (`comment.author === agentLogin`) would correctly skip the review — the positive E2E case is unreproducible by construction.

Per the issue's instructions ("If neither is available, you have to figure out how to test the agent-login disambiguation; ask back if blocked"), I am pausing the live E2E and asking for guidance. Suggested options:

1. Land #361 first; then I'll add the live E2E transcript as a follow-up comment here.
2. Cherry-pick #361's `--agent-login` plumbing into this branch (would expand scope past #366).
3. Stage a one-off temp shim on a throwaway branch just for the E2E run, ship without merging the shim.

Existing unit coverage proves the decision logic; the live E2E is purely about exercising the GitHub REST endpoints from the running daemon.

Refs #366

cc @bingran-you


bingran-you (changes requested):
[packages/github-scan/src/github-scan/engine/runtime/auto-revert.ts:160] and [:182] still stop at `per_page=100` and never pass `--paginate`, so a PR with more than 100 reviews or inline review comments can still stay stuck in `github-scan:human` if the qualifying human reply lands on a later page. Issue #366 explicitly called out pagination for these new surfaces, so this needs to be fixed or called out as an explicit follow-up dependency before approval. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

serenakeyitan:
## End-to-end verification — partial

Test was interrupted mid-run (agent process terminated unexpectedly), but the GitHub timeline + label-event evidence on the test target preserves the positive-case result.

### Positive case ✅ PASSED
- Test target: https://github.com/agent-team-foundation/first-tree/pull/351
- 02:18:27 UTC — applied `github-scan:human` to PR #351
- 02:18:44 UTC — posted PR **review** (NOT issue comment), 64 chars: *"Please proceed with this approach — confirmed via #366 E2E test."* — review ID 4214149312
- 02:19:15 UTC — daemon poll auto-stripped `github-scan:human` (verified via timeline event: `actor=serenakeyitan, event=unlabeled, label=github-scan:human` at the daemon's `--agent-login some-other-login` identity, so the strip was on the daemon's behalf)

This proves the new `fetchPrReviews` path triggers auto-revert on PR review bodies — the headline behaviour change.

### Negative case ⚠️ NOT VERIFIED
The agent applied the label a second time (02:19:47 UTC) and was killed before posting the empty-body APPROVE. So the empty-body-skip path was not exercised live. Unit test `tests/github-scan/github-scan-auto-revert.test.ts` covers this case (`PR review with empty body (APPROVE-only) → does NOT trigger`).

### Test pollution cleanup
- `github-scan:human` re-applied at 02:19:47 → removed manually
- The orphan test PR review (#pullrequestreview-4214149312) remains on PR #351 because GitHub's REST API does not permit dismissing `COMMENTED` reviews (only `APPROVED`/`CHANGES_REQUESTED`). Posted a clarifying comment on PR #351: https://github.com/agent-team-foundation/first-tree/pull/351#issuecomment-4362654050

### E2E setup (for reproducibility)
- Daemon: PR #369 build, port `:7879`, separate `GITHUB_SCAN_DIR`. Production daemon (port 7878) untouched.
- Agent-identity disambiguation via temp env-shim `GITHUB_SCAN_AGENT_LOGIN_OVERRIDE=some-other-login` since #361 hasn't merged. Shim reverted before this comment (worktree `git diff` clean).

cc @bingran-you

serenakeyitan:
Good catch — pushed 3aaa410 to address this.

**Fix:** all three fetchers (`fetchIssueComments`, `fetchPrReviews`, `fetchPrReviewComments`) now pass `--paginate` and decode the concatenated multi-page stdout via `splitConcatenatedJsonArrays`. The pre-existing `fetchIssueComments` had the same single-page limit, so it's fixed in the same commit.

**Why null-on-malformed-page:** if any page fails to JSON-parse, `decodePaginatedComments` returns `null` rather than a partial result. A partial result would risk masking a late-page qualifier (the exact failure mode you flagged), so we degrade to "fetch failed → don't strip" — which is already the contract the driver expects.

**Refactor:** `splitConcatenatedJsonArrays` moved from `commands/poll.ts` to `runtime/gh.ts` so a runtime module can import it without creating a runtime→commands cycle. `poll.ts` re-exports it so `github-scan-poll.test.ts` (which imports it from there) keeps working unchanged.

**Tests:** 5 new cases in `github-scan-auto-revert.test.ts` exercise `--paginate` + multi-page decoding for each fetcher, plus the null-on-malformed-page contract. Full suite: 460 + 77 = 537 pass, lint + typecheck clean.

bingran-you (changes requested):
[packages/github-scan/src/github-scan/engine/runtime/auto-revert.ts:232] `fetchHumanLabelAppliedAt()` is still capped at the first 100 timeline events. The issue timeline endpoint is oldest-first, so on a busy issue/PR the most recent `github-scan:human` label application can land on page 2+. In that case this function returns `null`, the driver emits `missing label-applied timestamp`, and the item never auto-reverts even though the comment/review fetchers are now paginated. Since #366's acceptance criteria says pagination is handled, I don't think we can clear the blocker until the timeline fetch is paginated too and covered by a regression test.

serenakeyitan:
Good catch — pushed dfb1dbf to address this.

**Fix:** `fetchHumanLabelAppliedAt` now passes `--paginate` and decodes the concatenated multi-page stdout via `splitConcatenatedJsonArrays`, matching the approach used by the three comment fetchers. The timeline endpoint is oldest-first, so without pagination the most recent `github-scan:human` label event on a busy PR (100+ timeline events) would silently land on a later page, causing the driver to emit `missing label-applied timestamp` and skip the auto-revert.

**Null-on-malformed-page:** same contract as the comment fetchers — if any page fails to JSON-parse, return `null` so the driver degrades to "don't strip" rather than acting on partial data.

**Tests:** 4 new cases in `github-scan-auto-revert.test.ts`:
1. Label on page 2 found via `--paginate`
2. Latest timestamp across pages when label applied multiple times
3. Malformed page returns `null`
4. Non-zero exit returns `null`

Full auto-revert suite: 28 tests pass. Lint + typecheck clean.

serenakeyitan:
Addresses your latest feedback. `fetchHumanLabelAppliedAt` now paginates through the timeline endpoint with the same 50-page hard cap as the comment / commit fetchers, tracking the LATEST qualifying `labeled` event across all pages. Added regression test covering page 2+ scenario.

Note: while this PR was open, main moved (PRs #381–#385 merged) and #370 already landed forward-pagination + `AUTO_REVERT_MAX_PAGES` for `fetchHumanLabelAppliedAt`. I merged main into this branch — keeping main's stricter pattern (forward page-loop with `console.warn` on cap) as the canonical implementation rather than the earlier `--paginate` style. The page-2 regression test you asked for is at `fetchHumanLabelAppliedAt — pagination (issue #365)` in `github-scan-auto-revert.test.ts`. I also extended the same forward-pagination pattern to the new PR-review and PR-review-comments fetchers introduced by this PR, with their own page-2 + hard-cap regression tests under `fetchPrReviews / fetchPrReviewComments — pagination (issue #366)`.

Tests: 502 pass in `@first-tree/github-scan` (52 in the auto-revert suite). Lint + typecheck clean.

cc @bingran-you

bingran-you (approved):
Re-reviewed the latest branch and I do not see remaining blockers. The forward-pagination path now covers the timeline plus both PR review surfaces, and the updated regressions cover the page-2 / cap cases I was concerned about. I also verified locally with `pnpm --filter @first-tree/github-scan exec vitest run tests/github-scan/github-scan-auto-revert.test.ts`, `pnpm --filter @first-tree/github-scan test`, `pnpm typecheck`, and `pnpm lint`. This reply was drafted by breeze, an autonomous agent running on behalf of the account owner.

serenakeyitan:
Thanks for the thorough re-review and approval @bingran-you! CI is green and all feedback has been addressed. Branch protection policy is currently blocking the merge — will need a maintainer to land this or an additional required approval. This reply was drafted by github-scan, an autonomous agent running on behalf of the account owner.

PR #370, opened 2026-05-02
fix(github-scan): paginate auto-revert timeline + comments fetches (#365)

serenakeyitan:
## Summary

`fetchHumanLabelAppliedAt` and `fetchIssueComments` in `auto-revert.ts` only fetched page 1 (`per_page=100`) of the GitHub REST endpoints. On long-lived issues with > 100 timeline events / comments the relevant signal could fall off page 1 and auto-revert silently failed — the human handoff stayed stuck because the label timestamp returned `null`.

This PR adds pagination with early-exit and a hard cap.

## Implementation

- **`fetchHumanLabelAppliedAt`** — forward pagination, accumulating the latest `labeled` event for `github-scan:human` across pages. Stops on a short or empty page.
- **`fetchIssueComments`** — paginates with `sort=created&direction=desc` so newest comments arrive first, then early-exits when a page's newest comment is at-or-before `labelAppliedAt` (later pages are older by definition and cannot trigger a revert). The driver now passes `labelAppliedAt` into the comments fetch.
- **Hard cap** of 50 pages (5,000 events) on both fetches, with a `console.warn` when the cap engages so a silent runaway can never happen.

## Test cases (10 new, all pass)

| Function                       | Case                                                                  |
| ------------------------------ | --------------------------------------------------------------------- |
| `fetchHumanLabelAppliedAt`     | 0-page (empty timeline) — returns `null`, 1 fetch                     |
| `fetchHumanLabelAppliedAt`     | 1-page (short page) — existing behaviour preserved                    |
| `fetchHumanLabelAppliedAt`     | 2-page — label event on page 2 is found                               |
| `fetchHumanLabelAppliedAt`     | hard-cap warning fires after 50 full pages                            |
| `fetchIssueComments`           | 0-page (empty) — returns `[]`                                         |
| `fetchIssueComments`           | 1-page (short page) — existing behaviour preserved                    |
| `fetchIssueComments`           | 3-page accumulation (no `labelAppliedAt`) — walks until short page    |
| `fetchIssueComments`           | early-exit fires — page 3 not fetched once page 2 is pre-label        |
| `fetchIssueComments`           | 3-page with post-label comment on page 3 (label older than every page) |
| `fetchIssueComments`           | hard-cap warning fires after 50 full pages                            |

All 11 existing auto-revert tests still pass; full suite: 457 tests pass in `@first-tree/github-scan`, 77 in `apps/cli`. `pnpm lint` and `pnpm typecheck` both clean.

## Verification

- [x] `pnpm -r test` — all green (534 tests)
- [x] `pnpm lint` — 0 warnings, 0 errors
- [x] `pnpm typecheck` — clean
- [ ] No live E2E (manufacturing > 100 timeline events is slow; mocked-page unit tests are the right level — confirmed with the issue author)

Refs #365
cc @bingran-you


bingran-you (approved):
Looks good — small, targeted fix for a real failure mode, the early-exit on desc-sorted comments is the right shape, and the test matrix is thorough (0/1/many pages, early-exit assertion, hard-cap warn). Approving with a few nits worth a follow-up but not blockers.

**Nit 1 — warn message in `fetchHumanLabelAppliedAt` is backwards.** The `/issues/{n}/timeline` endpoint returns events oldest-first (ascending). When the 50-page cap engages on a >5,000-event thread, we're walking the *first* 5,000 events and missing everything after — i.e., the **most recent** history is clipped, not "earlier history". The message at `packages/github-scan/src/github-scan/engine/runtime/auto-revert.ts:277` should say something like `most-recent history clipped` (and operationally that's the worse failure mode here, since a more recent re-label would be the one we'd want).

**Nit 2 — pre-label page is pushed into `all` before early-exit.** At `auto-revert.ts:168-176`, `all.push(...pageComments)` runs before the `firstMs <= labelMs` check, so the entire pre-label page gets accumulated and then filtered downstream by Guard 4 in `shouldAutoRevertHuman`. Harmless (the decision is still correct), just a tiny waste — moving the early-exit check above the push would skip it.

**Nit 3 — double JSON parse per timeline page.** `fetchHumanLabelAppliedAt` parses `result.stdout` once at lines 255-261 to get `pageLen`, then `parseTimelinePage(result.stdout)` parses it again. One pass that returns both the length and the hits would be cleaner.

None of these change behavior on the realistic paths the tests cover, so happy to land as-is and tackle in a follow-up if you'd rather keep this PR scoped to #365.

serenakeyitan:
Thanks for the careful review @bingran-you — all three nits are real and worth fixing now rather than as a follow-up. Pushed e810013:

**Nit 1 — backwards warn message.** You're right, `/issues/{n}/timeline` is oldest-first, so the cap clips the *most recent* history (and that's the operationally worse failure mode). Updated the message to `most-recent history may be clipped` and the docstring on `fetchHumanLabelAppliedAt` now spells out the asc-order reasoning.

**Nit 2 — pre-label page accumulated before early-exit.** Moved the early-exit check above `all.push(...)` in `fetchIssueComments` so an entirely-pre-label page is no longer accumulated just to be filtered downstream by Guard 4. Updated the early-exit test: it now asserts `length === PAGE_SIZE * 1` (page 1 only) instead of `* 2`.

**Nit 3 — double JSON parse.** Collapsed `parseTimelinePage` to return both the label hits and the page length in a single pass; `fetchHumanLabelAppliedAt` no longer re-parses `result.stdout`.

All 457 tests pass, lint + typecheck clean.

Issue #382, opened 2026-05-03
fix(github-scan): drop `AUTO_REVERT_MIN_BODY_CHARS` guard — blocks LGTM and short Chinese replies


serenakeyitan:
## Problem

`auto-revert.ts` filters qualifying human comments with a hardcoded `> 20 chars (trimmed)` guard, originally added in #358 to avoid spurious re-dispatch on tiny acks like 👍 or `"ok"`. In practice this guard is too aggressive and creates real false negatives:

- **English short approvals are blocked.** `LGTM` (4 chars), `go ahead` (8), `approved` (8), `LGTM, ship it` (13) — all standard code-review approve signals. Author replies "LGTM, please proceed", daemon ignores it, PR stays stuck in `github-scan:human`.
- **Cross-language inconsistency.** 20 ASCII chars ≈ 4-5 English words ≈ a single short sentence. 20 Chinese chars ≈ 20 汉字 ≈ a paragraph. A Chinese user has to write substantially more to trigger the same auto-revert path. `"请继续推进"` (5 汉字) is a complete, intentional approve — silently dropped.
- **Hardcoded, not configurable.** Operators can't tune it for their team's writing style.

The guard's stated goal — *"don't re-fire on emoji acks"* — is already covered by the fact that **reactions are not comments** in GitHub's API. Reactions go through `/issues/{n}/reactions`, not `/issues/{n}/comments`. `fetchIssueComments` never sees them.

So the guard is solving a problem that doesn't exist in practice, while creating real false negatives.

## Proposal

**Drop the body-length guard entirely.** Keep the two essential guards:

1. `comment.user.login !== agentLogin` — prevents the agent from re-triggering on its own comments.
2. `comment.created_at > labelAppliedAt` — prevents pre-label comments from triggering on stale state.

These two are sufficient. The trade-off:

- ✅ `LGTM`, `go ahead`, `请继续`, etc. now trigger — correct.
- ⚠️ Tiny reactive replies (`"ok"`, `"thx"`) will trigger one re-dispatch each. The agent looks at the PR, sees there's nothing to act on, re-applies `github-scan:human`. One wasted dispatch cycle, ~30s. Acceptable cost — much smaller than the current "wait forever" failure mode.

## Acceptance criteria

- [ ] `AUTO_REVERT_MIN_BODY_CHARS` removed from `auto-revert.ts` (the constant and its usage).
[...]

PR #384, opened 2026-05-03
fix(github-scan): drop AUTO_REVERT_MIN_BODY_CHARS guard (#382)

serenakeyitan:
## Summary

Drops the hardcoded `body.trim().length > 20` filter from `shouldAutoRevertHuman` in `packages/github-scan/src/github-scan/engine/runtime/auto-revert.ts`. That guard was producing real false negatives for legitimate short approvals — `LGTM` (4 chars), `go ahead` (8), `请继续推进` (5 汉字) — which left PRs stuck in `github-scan:human` even after the human had explicitly said "go".

The original rationale for the guard ("filter emoji acks") is moot: GitHub reactions are not comments in the REST API and never reach `fetchIssueComments`, so the guard was solving a problem that doesn't exist while creating one that does.

The two essential guards remain:
1. `comment.user.login !== agentLogin` — own-comment filter.
2. `comment.created_at > labelAppliedAt` — strictly-after-label filter.

Trade-off: tiny reactive replies like `ok`/`thx` will now trigger one wasted re-dispatch (~30s) before the agent re-applies `github-scan:human`. Acceptable cost vs. the current "wait forever" failure mode.

## Test changes

- Updated `(b) short ack` case: now asserts a short post-label reply **does** trigger revert (was: does not).
- Added `(b1)`: `LGTM` from non-agent user, post-label → triggers revert.
- Added `(b2)`: `请继续推进` (5 Chinese chars) from non-agent user, post-label → triggers revert.
- Updated module/test docstrings to describe the new three-guard list.
- Existing author-guard, case-insensitivity, timestamp-guard, pre-label, pagination, and driver-end-to-end tests unchanged and still pass.

## Verification

- `pnpm --filter github-scan test` → 35 files / 475 tests passed.
- `pnpm lint` → clean.
- `pnpm typecheck` → clean.
- `pnpm -r test` → green (note: the unrelated `github-scan-statusline` cold-start perf test is flaky under load and occasionally trips its 200ms cap; passes consistently in isolation).

Per the issue, no live E2E was required — this is a pure unit-level removal of a gate; behaviour is identical to the existing tested code paths just without the char filter.

## Test plan

- [x] Unit tests pass (`pnpm --filter github-scan test`)
- [x] Lint + typecheck clean
- [x] Existing acceptance tests for #358 still pass (own-comment, reaction-empty, pre-label, strict-inequality, case-insensitive)

Refs #382
cc @bingran-you


bingran-you (approved):
No blocking findings. The behavior change matches issue #382, and the added coverage exercises the new short-reply cases directly.

I couldn't rerun the package tests in this workspace because dependencies are not installed here (`tsdown: command not found`), so I'm relying on the green CI checks for execution verification.

serenakeyitan:
Thanks for the review @bingran-you! CI is green and we have approval — ready to merge. The branch protection policy is currently blocking the merge, so this will need an admin merge or an additional required approval to land.

Issue #395, opened 2026-05-03
fix(cli): suppress react-devtools-core unresolved-import warning during build


serenakeyitan:
## Problem

Every \`pnpm --filter first-tree build\` prints:

\`\`\`
[UNRESOLVED_IMPORT] Warning: Could not resolve 'react-devtools-core' in
.../node_modules/.pnpm/ink@7.0.1_.../node_modules/ink/build/devtools.js
Module not found, treating it as an external dependency
\`\`\`

\`react-devtools-core\` is a development-only optional dep of \`ink\`. The bundler treats it as external (correct), but the warning is alarming to anyone watching install/build output — especially during \`first-tree github scan install\` where it scrolls past mid-flow.

## Proposal

Configure tsdown / rolldown in \`apps/cli/\` to either:
1. Mark \`react-devtools-core\` as a known external (silences the warning), OR
2. Suppress unresolved-import warnings for known-optional deps.

## Refs

Surfaced by E2E onboarding test for the menu-bar tray on \`feat/tray\` branch.

cc @bingran-you

serenakeyitan:
Tracked under design proposal #408 / implementation PR #409.

serenakeyitan:
Implemented and shipped via #396.

PR #396, opened 2026-05-03
fix(cli): mark react-devtools-core as external to suppress build warning (#395)

serenakeyitan:
## Summary

Rolldown was emitting an `[UNRESOLVED_IMPORT]` warning on every CLI build:

```
ink/build/devtools.js (7:21) Warning: Could not resolve 'react-devtools-core'
```

`react-devtools-core` is an optional dev-only dep of `ink` that is never reachable at runtime in our CLI. The cleanest fix is to mark it as a known external in the bundler config so rolldown stops warning about it.

## Change

`apps/cli/tsdown.config.ts` — add `"react-devtools-core"` to the main bundle's `external` array.

## Verification

- `pnpm --filter first-tree build` — no longer prints the `UNRESOLVED_IMPORT` warning.
- `node apps/cli/dist/index.js --version` — returns `0.4.0-alpha.1`.

Refs #395

cc @bingran-you


bingran-you (approved):
LGTM. Small, surgical fix that does exactly what issue #395 proposes (option 1): marking `react-devtools-core` as a known external in the main CLI bundle so rolldown stops emitting the `UNRESOLVED_IMPORT` warning. The inline comment explains the reasoning well, and scoping the change to just the main bundle (not the statusline entry, which doesn't pull in ink) is the right call.

PR #428, opened 2026-05-15
feat: add onboarding rule-layer automation

bingran-you:
## Summary
- add versioned rule-layer workflow template management and install Tier 0 `validate.yml` during bootstrap/upgrade
- add `first-tree tree automation install --tier 2` to stage the owners-gate rollout without executing repo-policy `gh api` commands on the user's behalf
- update onboarding skill/docs plus local/live e2e coverage, including the prompt-driven `mews` onboarding smoke and a live GitHub `validate` workflow check

## Testing
- `pnpm ci:check`
- `RUN_CODEX_PROMPT_SMOKE=0 RUN_CLAUDE_PROMPT_SMOKE=0 ./e2e/onboarding-smoke.sh`
- `RUN_CODEX_PROMPT_SMOKE=1 RUN_CLAUDE_PROMPT_SMOKE=0 AGENT_TIMEOUT_SECS=420 ./e2e/onboarding-smoke.sh`
- `TREE_REPO_VISIBILITY=private ./e2e/github-validate-e2e.sh`

## Notes
- The prompt-driven onboarding smoke uses: `Use the latest First-Tree CLI to install the skill in the current repository and complete the onboarding process by running /first-tree-onboarding skill and build an initial version of tree using /first-tree-sync skill`.
- The successful live validate run used a temporary tree repo under `bingran-you/mews-tree-e2e-1778807032`.


apps/cli/src/commands/tree/bootstrap.ts
```diff
@@ -115,6 +116,7 @@ export function bootstrapTreeRoot(
   writeIfMissing(join(targetRoot, ".first-tree", "org.yaml"), renderOrgConfigPlaceholder());
   writeIfMissing(join(targetRoot, TREE_VERSION_FILE), "0.4.0-alpha.1");
   writeIfMissing(join(targetRoot, TREE_PROGRESS_FILE), renderTreeProgress());
+  ensureTier0RuleLayer(targetRoot);
```
bingran-you: `ensureTier0RuleLayer` returns a `Tier0RuleLayerSummary` whose `validate.kind` can be `written`, `skipped-same-version`, `skipped-existing-no-marker`, or `needs-upgrade`. Here the result is discarded, and line 162 unconditionally prints `Ensured <validateWorkflowPath>`.

For the proposal's §8 template policy to actually reach the user, the bootstrap output needs to distinguish:
- `written` → `Ensured ...`
- `skipped-same-version` → silent / `Ensured (up to date) ...`
- `skipped-existing-no-marker` → `Skipped <path> (custom file, no managed marker)`
- `needs-upgrade` → `Upgrade available: <path> is at v<N>, bundled v<N+1>. Run \`first-tree tree upgrade\` to update.`

Right now a user with a v0 managed `validate.yml` would see "Ensured" on every bootstrap and never learn they need to upgrade. (Also applies to `upgrade.ts:82`.)

apps/cli/src/commands/tree/upgrade.ts
```diff
@@ -78,6 +79,7 @@ function upgradeTreeRoot(targetRoot: string, bundledSkillVersion: string): Upgra
   ensureWhitepaperSymlink(targetRoot);
   upsertLocalTreeGitIgnore(targetRoot);
   writeFileSync(join(targetRoot, TREE_VERSION_FILE), `${bundledSkillVersion}\n`);
+  ensureTier0RuleLayer(targetRoot);
```
bingran-you: Same issue as in `bootstrap.ts:119` — the `Tier0RuleLayerSummary` is discarded. This one is the bigger surprise: a user who runs `first-tree tree upgrade` precisely *because* they want a newer template will get a silent `needs-upgrade` result with no on-screen indication. Either:

1. Surface `needs-upgrade` and `skipped-existing-no-marker` in the upgrade summary; or
2. On upgrade, treat `needs-upgrade` as a write (the user explicitly asked to upgrade) — but then keep a backup of the old file, per proposal §8 ("only prompt for an upgrade," not silent overwrite).

I'd lean toward option 1 plus a follow-up `first-tree tree automation upgrade --apply` flag for the actual overwrite, so the upgrade command stays no-clobber by default.

apps/cli/src/commands/tree/tree-templates.ts
```diff
@@ -125,6 +125,562 @@ export function renderOrgConfigPlaceholder(): string {
...
+    "    runs-on: ubuntu-latest",
+    "    steps:",
+    "      - uses: actions/checkout@v4",
+    "",
+    "      - uses: actions/setup-node@v4",
+    "        with:",
+    '          node-version: "22"',
+    "",
+    "      - name: Validate tree",
+    "        run: npx -p first-tree first-tree tree verify",
```
bingran-you: `npx -p first-tree first-tree tree verify` pulls the latest published CLI on every PR. If a future CLI release breaks `tree verify`'s exit code or output contract, every onboarded tree's `validate` check breaks simultaneously — and these are the gating status check the proposal wants every tree to require.

Two options:
1. Pin to a minor version: `npx -p first-tree@^0.4 first-tree tree verify`. Couples the workflow template to a CLI major line; you bump the template version when you bump the line.
2. Inject the bundled CLI version at render time, e.g. `npx -p first-tree@${readBundledSkillVersion()} ...`. Stronger pin, but requires re-running `tree upgrade` to track CLI releases.

Either way, the workflow needs *some* version anchor before this becomes the default Tier 0 gate.

apps/cli/src/commands/tree/automation.ts
```diff
@@ -0,0 +1,643 @@
...
+
+Options:
+  --tier <number>     automation tier to install (currently only 2 is supported)
+  --tree-path PATH    operate on a tree repo from another working directory
+  --dry-run           inspect state and print next steps without writing files
+  --help              show this help message`;
+
+function configureInstallAutomationCommand(command: Command): void {
+  command
+    .requiredOption("--tier <number>", "automation tier to install (currently only 2 is supported)")
```
bingran-you: `requiredOption` plus the runtime check `if (options.tier !== "2")` makes the UX awkward: the user must pass `--tier`, but the only accepted value is `2`. Two cleaner shapes:

1. **Default to `2`**: `option("--tier <number>", "...", "2")` — `first-tree tree automation install` just works.
2. **Subcommand per tier**: `first-tree tree automation install tier2` — explicit, leaves room for tier-3 etc.

Option 1 is the smallest delta and gets us out of the "required flag with one value" footgun without breaking the existing call sites in the e2e scripts (which already pass `--tier 2`).

apps/cli/src/commands/tree/automation.ts
```diff
@@ -0,0 +1,643 @@
...
+
+type RemoteWorkflowState = "custom" | "current" | "missing" | "outdated";
+
+type WorkflowDefinition = {
+  name: "auto-merge" | "review-enforcer";
+  path: string;
+  templateVersion: number;
+};
+
+export const AUTOMATION_INSTALL_USAGE = `usage: first-tree tree automation install --tier 2 [--tree-path PATH] [--dry-run]
```
bingran-you: `AUTOMATION_INSTALL_USAGE` is exported but I don't see it referenced anywhere — the `configure` callback uses commander's own option strings, and no test/help path imports this constant. Either wire it into a `--help` override, or drop the export.

bingran-you (commented):
Reviewed against the v7 proposal `tree-repo-github-automation-onboarding.20260512.md`.

## Alignment with the proposal

The rule / agent split is faithfully implemented:

- **§1 "Rules in, agents out"** — every workflow this PR ships (`validate.yml`, `auto-merge.yml`, `review-enforcer.yml`) is rule-only. No model calls, no provider keys, no in-repo agent runner. ✅
- **§2 three tiers** — Tier 0 ships from `ensureTier0RuleLayer` (bootstrap + upgrade), Tier 1 is explicitly not installed, Tier 2 is gated behind `tree automation install --tier 2`. ✅
- **§3 default flow** — bootstrap/upgrade write only the Tier 0 rule layer; no secrets, no GitHub App install, no ruleset change. ✅
- **§7 hard rules** — secrets never enter chat (none asked for); ruleset-changing `gh api` commands are printed, never executed (`nextCommands` array + the new "Never execute the Tier 2 ruleset `gh api` commands yourself" hard rule in SKILL.md); idempotency is derived from live `gh api` state, no skill-side progress file. ✅
- **§8 template policy** — `# first-tree-template-version: <n>` marker is on every template; `writeTemplatedFile` handles missing / no-marker / same-version / older-version correctly; older managed templates return `needs-upgrade` rather than being overwritten. ✅ (one caller swallows this — see inline comments)
- **§9 Tier 2 activation order** — the four stages (`write_rule_layer` → `create_ruleset` → `activate_ruleset` → `configured`) map directly onto the proposal's three-step rollout, and the ruleset is created with `enforcement: "evaluate"`. ✅
- **§10 target state compatibility** — nothing in this PR ties onboarding to the transitional "workflow + model API key" shape; when Hub dispatch lands, this skill won't need to be rewritten. ✅

## Open questions / follow-ups

1. **`first-tree-hub` is a private repo** — the user-facing onboarding summary links to it as the place to enable Tier 1, but external users will hit a 404. See inline.
2. **`first-tree-gate` GitHub App** — the App install URL is referenced from both `auto-merge.yml` comments and the CLI warning. The URL resolves, but worth confirming the app is the intended one and is reachable for typical users. See inline.
3. **Hub-side discovery is hardcoded** — proposal §10 pre-condition is to "detect whether the org already has the GitHub App installed and Hub dispatch configured, so that the post-init message points at the right place." The PR hardcodes a single link instead. Probably fine as a v1 and worth a follow-up rather than blocking this PR.
4. **Duplicate ownership logic** between `auto-merge.yml` and `review-enforcer.yml` (~80 lines of inline github-script in each). Proposal §5 already flags this as a future refactor of the rule layer (extract into a CLI command). Non-blocking.

## Inline comments below cover

- Discarded `TemplateWriteResult` in `bootstrap.ts` and `upgrade.ts` — `needs-upgrade` and `skipped-existing-no-marker` should surface, not be silenced.
- Unpinned `npx -p first-tree` in `validate.yml` — every tree's CI is on latest-CLI.
- `requiredOption("--tier")` UX when only `"2"` is accepted.
- Dead-looking `AUTOMATION_INSTALL_USAGE` export.
- `first-tree-hub` private-repo link in the success summary.

Overall this matches the proposal closely and the staged-rollout machinery is well-tested. Most comments are polish, not blockers.

apps/cli/src/commands/tree/automation.ts
```diff
@@ -0,0 +1,643 @@
...
+      tier: options.tier,
+      warnings,
+      workflowFiles,
+    };
+  }
+
+  const appInstalled = detectAppInstallation(targetRoot, repoMetadata.repoSlug, runner);
+  if (appInstalled === false) {
+    warnings.push(
+      `The \`first-tree-gate\` GitHub App does not appear to be installed on ${repoMetadata.repoSlug}. Install it before turning Tier 2 on: ${FIRST_TREE_GATE_INSTALL_URL}`,
```
bingran-you: Two follow-ups on this warning:

1. `detectAppInstallation` uses `repos/<slug>/installation`, which only returns 200 when the *authenticated caller* can see the App's installation on this repo. With a token scoped without that access (a common case for org members who didn't install the App themselves), this returns 404 and we'd print a false-negative "App not installed" warning. Worth a one-line caveat in the warning text, or detect-via-org-installations as a fallback.

2. Confirm `https://github.com/apps/first-tree-gate` is the canonical install URL for the gate App you want users to install. The URL resolves to a 200 page, but if it's a stub / wrong owner / placeholder, every onboarded tree's user gets pointed at the wrong place.

skills/first-tree-onboarding/SKILL.md
```diff
@@ -157,6 +184,9 @@ If none match → skip Phase C. The combined check is reliable because all three
    Tree repo:   <tree_root> (<bindingMode>, <treeMode>)
    Tree URL:    <treeRemoteUrl or "not published">
    Daemon:      <running for <owner/repo> | skipped>
+   GitHub Actions: validate.yml installed (Tier 0, rule-based)
+   AI PR review:  not installed by this skill. Enable via first-tree-hub: https://github.com/agent-team-foundation/first-tree-hub
```
bingran-you: `https://github.com/agent-team-foundation/first-tree-hub` is a **private** repo (just verified via the GitHub API). External users following the success summary will hit a 404 here, which is the worst possible UX for the message that's supposed to explain where Tier 1 lives.

Three options:
1. Drop the link until `first-tree-hub` has a public onboarding doc — say `"AI PR review: not installed by this skill. Owned by first-tree-hub (ask your org admin to enable)."`
2. Link to a public doc/anchor under `first-tree` itself that explains the Hub split (could live in this PR alongside `references/github-automation.md`).
3. Make `first-tree-hub` public, or at least publish a `README` / docs site whose URL resolves for everyone.

The proposal §10 pre-condition ("Define how `first-tree-onboarding` detects whether the org already has the GitHub App installed and Hub dispatch configured, so that the post-init message points at the right place") points at option 2/3 as the eventual answer — this comment is just flagging that the current hardcoded link breaks for anyone outside the org.

apps/cli/src/commands/tree/tree-templates.ts
```diff
@@ -125,6 +125,562 @@ export function renderOrgConfigPlaceholder(): string {
...
+  ].join("\n");
+}
+
+export function renderAutoMergeWorkflow(): string {
+  return [
+    "# first-tree-template-version: 1",
```
bingran-you, line 157: Heads-up / non-blocking: `renderAutoMergeWorkflow` and `renderReviewEnforcerWorkflow` share ~80 lines of inline github-script (`parseInlineOwners`, all-members enumeration, `readBaseContent`, `parseOwnersField`, owner-set intersection). If the ownership-rule semantics ever change (e.g. `owners: [team:foo]`, group expansion, anything past `[*]`), both workflows have to be updated in lockstep.

Proposal §5 already flags this as a future refactor (extract into a `first-tree` CLI command so the workflow templates become thin shims). Worth tracking as a follow-up issue; not a blocker for this PR.

